PR #1, opened 2019-08-19
PR for ItemDetailList.js refactoring

utku-ozturk:
1. static Detail.formKey and Detail.formValue functions converted to DetailRowLabel and DetailRowValue components respectively.
2. TabErrorBoundary added to handle exceptions in tabs. (is to prevent page total failure, instead only the related tab will fail. Recently, only used for Details tab.)
3. static functions moved from components to global scope (static shouldUseTable is not changed, since the validation may break down. We may use a Json-query library, e.g. https://github.com/s3u/JSONPath, to simplify json traverse and validation. If make sense, we can dig it deeper.) 
4. some PropTypes and documentation added

PS: 1. we cherry-picked your (@alexkb0009 ) commits after the branch created. Instead of cherry-pick should we merge or rebase, not decided?

2. ItemDetailList component structure:
ItemDetailList

### DIFF
--- a/src/components/browse/SearchView.js
+++ b/src/components/browse/SearchView.js
@@ -176,8 +176,8 @@ class ControlsAndResults extends React.PureComponent {
     }
 
     forceUpdateOnSelf(){
-        var searchResultTable   = this.searchResultTableRef.current,
-            dimContainer        = searchResultTable && searchResultTable.getDimensionContainer();
+        const searchResultTable = this.searchResultTableRef.current;
+        const dimContainer = searchResultTable && searchResultTable.getDimensionContainer();
         return dimContainer && dimContainer.resetWidths();
     }
 
@@ -287,11 +287,9 @@ export class SearchView extends React.PureComponent {
     render() {
         const { facets : propFacets, navigate: propNavigate, context } = this.props;
         return (
-            <div className="container" id="content">
-                <div className="search-page-container">
-                    <AboveSearchTablePanel {..._.pick(this.props, 'href', 'context', 'schemas')} />
-                    <SearchControllersContainer {...this.props} facets={propFacets || context.facets} navigate={propNavigate || navigate} />
-                </div>
+            <div className="search-page-container">
+                <AboveSearchTablePanel {..._.pick(this.props, 'href', 'context', 'schemas')} />
+                <SearchControllersContainer {...this.props} facets={propFacets || context.facets} navigate={propNavigate || navigate} />
             </div>
         );
     }
--- a/src/components/browse/components/above-table-controls/RightButtonsSection.js
+++ b/src/components/browse/components/above-table-controls/RightButtonsSection.js
@@ -8,11 +8,12 @@ import { SearchResultTable } from './../SearchResultTable';
 
 
 export const RightButtonsSection = React.memo(function RightButtonsSection(props){
-    const { currentOpenPanel, onColumnsBtnClick } = props;
+    const { currentOpenPanel, onColumnsBtnClick, windowWidth, isFullscreen, toggleFullScreen } = props;
+    const showToggleLayoutBtn = (typeof windowWidth === 'number' && typeof isFullscreen === 'boolean' && typeof toggleFullScreen === 'function');
     return (
         <div className="right-buttons">
             <ConfigureVisibleColumnsButton onClick={onColumnsBtnClick} open={currentOpenPanel === "customColumns"} />
-            <ToggleLayoutButton {..._.pick(props, 'windowWidth', 'isFullscreen', 'toggleFullScreen')} />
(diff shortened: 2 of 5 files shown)


utku-ozturk:
Thanks for the detail review @alexkb0009. We are happy to be able to do something good, albeit minor in our first PR. But also we are aware that some tasks are incomplete as you stated. We'll ask more questions when get confused, and provide a good description for our assumptions.

**About Item 1 & 2:**
 
We'll write in detail when we fix the items in https://github.com/4dn-dcic/shared-portal-components/pull/1#pullrequestreview-277252992, but we wanted to write some ideas for items 1&2 may help for a quick decision:

origin/master's ItemDetailList.js/DetailRow's render() function has a section as:
```javascript
let value = Detail.formValue(item, popLink, this.props['data-key'], itemType, columnDefinitions, 0, schemas, termTransformFxn);
...
if (value.type === SubItemTitle) {
```

static DetailRow.formValue function generates a code snippet, then checks its `type` and if-else blocks returns the actual code blocks.

When we refactored the code, we created `DetailRowValue` _function_ component instead of static `DetailRow.formValue` function as adviced in [https://reactjs.org/docs/thinking-in-react.html](url). This fix, unfortunately generated a new problem. We cannot get the type of component, since component not rendered yet. `TestRenderer` [https://en.reactjs.org/docs/test-renderer.html](url) comes as a life saver to convert components to pure js objects without DOM. Of course, it is not a pretty solution. (Besides its weird name in an production environment, it comes with its burden: an extra file in bundle.js.) We finally can get the `DetailRowValue`'s return type by the help of `TestRenderer`. But if a function component is memoized by React.memo, TestRenderer cannot get the correct type. (this may be related with the issue but not sure: [https://github.com/facebook/react-devtools/issues/1193](url)) So we converted the function components to class components including existing `SubItemTitle`. This is why we used class components instead of function components.

Alternatively, we can implement a function that gets `DetailRowValue`'s return type, honestly cannot decide: go back to the static formValue function with a new name, isn't it?

alexkb0009:
Thx, makes sense!

This potentially indicates that might be good idea to re-architect some aspects of this logic, but the only idea that have so far is similar to what you're saying and yeah agree isn’t much cleaner; not too sure if would be worth doing:

Could take approach like `shouldUseTable` function and then create 2 funcs which might resemble:
```javascript
function isCollapsibleObject(item){
    return ..is object, not null, not direct link to Item..
}

function isListOfCollapsibleObjects(item){
    return Array.isArray(item) && _.all(item, isCollapsibleObject) && !shouldUseTable(item);
}
```

...wherein both funcs could be called from both DetailRowValue (e.g. `isListOfCollapsibleObjects ` in place / inverse-if of where `shouldUseTable(item, ...)` is called, `isCollapsibleObject` in block/area/around where `const linkElement = itemUtil.generateLink(...` logic is) as well as in DetailRow (`isListOfCollapsibleObjects` in place of `valueInstance.findAllByType(SubItemTitle)` & `isCollapsibleObject` in place of `valueInstance.children[0].type === SubItemTitle`).

alexkb0009:
I've just performed 4.II from my first response (moving TabErrorBoundary to 4DN portal) and found an issue that should be accounted for in future --

The TabErrorBoundary as implemented originally requires a `props.href` and a `props.canonical` in order to send proper analytics event as well as to reset itself (upon URL change). None of these are being passed into it in original implementation:



Since this error boundary (as implemented) is only applicable to contents of this 1 tab, it becomes unmounted once change to a different tab, so theoretically the `componentDidUpdate` function could be removed completely. For `componentDidCatch` to keep sending page href to analytics, could just pass in `href={href}` (`href` is available in the props being passed into ItemDetailList component) (and maybe adjust error msg to inform that is occurring on a tab).


----

Having said that, [have moved the TabErrorBoundary over to 4DN portal's TabbedView.js file](https://github.com/4dn-dcic/fourfront/commit/82aaeb69db214f917355cb823a86b67ca674f5dd) and using `tabKey` now to inform of tab change / reset error state in place of `props.canonical` and attempting to grab `href` from browser window.

alexkb0009:
Heh, previous comment also helped to realize that href wasn't being passed to ContentErrorBoundary in App.js either [which just added small patch for](https://github.com/4dn-dcic/fourfront/commit/d0af6211777b4c4efa11b565bf38b0291d0a0570).

utku-ozturk:
> Heh, previous comment also helped to realize that href wasn't being passed to ContentErrorBoundary in App.js either [which just added small patch for](https://github.com/4dn-dcic/fourfront/commit/d0af6211777b4c4efa11b565bf38b0291d0a0570).

@alexkb0009  Great! 😄  our two negatives (`href`, `canonical`), at least, did something positive.

PR #6, opened 2019-09-30
Single Item Selection & SubmissionView Changes

utku-ozturk:
**Note:** this PR should be handled with PR https://github.com/4dn-dcic/fourfront/pull/1174

1. Single selection's UI is changed, so both `multi select` and `single select` popup UIs are alike.
2. `SubmissionView` is updated to open `multi` or `single selection` popup according to the field type.

Screen Shot 2019-09-30 at 17 42 53

### DIFF
--- a/src/components/forms/components/submission-fields.js
+++ b/src/components/forms/components/submission-fields.js
@@ -528,22 +528,22 @@ class LinkedObj extends React.PureComponent {
     }
 
     childWindowAlert(){
-        return null;
-        // const { schema, nestedField } = this.props;
-        // const itemType = schema && schema.linkTo;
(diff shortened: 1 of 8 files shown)


src/components/forms/components/submission-fields.js
```diff
@@ -528,21 +528,27 @@ class LinkedObj extends React.PureComponent {
     }
 
     childWindowAlert(){
-        const { schema, nestedField } = this.props;
+        const { schema, nestedField, isMultiSelect } = this.props;
         const itemType = schema && schema.linkTo;
         const prettyTitle = schema && ((schema.parentSchema && schema.parentSchema.title) || schema.title);
+        const message = (
+            <div>
+                { isMultiSelect?
```
alexkb0009: oops I needa inverse this check

PR #7, opened 2019-10-02
Facet's '!=' Condition Rendering Fixed

utku-ozturk:
**Note:** this PR should be handled with PR https://github.com/4dn-dcic/fourfront/pull/1176

**Problem**: Facets, of both `Browse` and `Search` pages, not rendering `!=` conditions correctly in UI, on the other hand backend works good and returns the correct query results.

**Solution**: Facet components only handle 2 states for terms: `selected` and `untouched`. But `!=` case also requires the 3rd state: `omitted`. Facets components are updated to handle those 3 states: terms having != condition, rendered as brick background color and minus icon on the left. (and common properties section on the bottom is also updated.)

**Future Work**: != condition can only be tested by tweaking URL on address bar. Any UI component not implemented yet, end-users hardly try it unless implemented soon.

New:
Screen Shot 2019-10-03 at 00 24 48

Old:
Screen Shot 2019-10-03 at 00 46 43


src/components/util/search-filters.js
```diff
@@ -246,6 +355,7 @@ export function getDateHistogramIntervalFromFacet(facet){
 
 
 /**
+ * @deprecated
```
alexkb0009: Might be safe to delete the deprecated functions. Maybe if have time can double-check via Search that these functions are not used anymore in CGAP, 4DN, and SPC, and then delete + push to master here.

utku-ozturk: Sure! we'll check once more and delete if not referenced anymore.

PR #11, opened 2019-11-07
Facet grouping

Bianca-Morris:
ad97a069c9f6fc3a8b42545c00a7ef33

Not sure if we should wait to merge this in or not; have the final schema decisions for grouping been officially made? Definitely needs testing on 4DN either way. And need to pull in changes from master (I'm having trouble getting new master on shared-portal-components working on my local).

Also, please take a look what I did with `mergeTerms()` and `anyTermsSelected()` in `index.js`. I pulled those methods out of `FacetTermsList` to make the indicator color and selected terms tooltips work in `FacetOfFacets`. Not sure if I should rework `FacetTermsList` to work the same way or not.

### DIFF
--- a/src/components/browse/components/FacetList/FacetOfFacets.js
+++ b/src/components/browse/components/FacetList/FacetOfFacets.js
@@ -2,11 +2,9 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import memoize from 'memoize-one';
 
 import { Collapse } from './../../../ui/Collapse';
 import { Fade } from './../../../ui/Fade';
-import { anyTermsSelected as anyTermsSelectedPerFacet, mergeTerms }  from './index';
 
 /**
  * Used to render individual facet fields and their available terms in FacetList.
@@ -20,16 +18,10 @@ export class FacetOfFacets extends React.PureComponent {
         super(props);
         this.handleOpenToggleClick = this.handleOpenToggleClick.bind(this);
         this.handleExpandListToggleClick = this.handleExpandListToggleClick.bind(this);
-        this.checkAllTerms = this.checkAllTerms.bind(this);
         this.state = {
             'facetOpen'     : typeof props.defaultFacetOpen === 'boolean' ? props.defaultFacetOpen : true,
             'facetClosing'  : false,
             'expanded'      : false
-        };
-        this.memoized = {
-            checkAllTerms: memoize(this.checkAllTerms),
-            anyTermsSelectedPerFacet: memoize(anyTermsSelectedPerFacet),
-            mergeTerms: memoize(mergeTerms)
         };
     }
 
@@ -61,28 +53,10 @@ export class FacetOfFacets extends React.PureComponent {
         });
     }
 
-    checkAllTerms() {
-        const { facets = [], filters = [] } = this.props;
-        let anySelected = false;
-        // console.log("log1: running CheckAllTerms on facetList: ", facets);
-        facets.forEach((facet) => {
-            // console.log("log1: examining this facet for selected terms: ", facet.props.facet);
-            // console.log("log1: seeking match with these filters", filters);
-            const terms = this.memoized.mergeTerms(facet.props.facet, filters);
-            // console.log("log1: merged terms, ", terms);
-            const anyTermsSelectedThisFacet = this.memoized.anyTermsSelectedPerFacet(terms, facet.props.facet, filters);
-            if (anyTermsSelectedThisFacet === true) {
-                anySelected = true;
-            }
-        });
-        return anySelected;
-    }
 
     render() {
-        const { title, facets, tooltip } = this.props;
+        const { title, facets, tooltip, areTermsSelected } = this.props;
         const { facetOpen, facetClosing } = this.state;
-
-        const anyTermsSelected = this.memoized.checkAllTerms();
 
         return (
             <div className={"facet" + (facetOpen ? ' open' : ' closed') + (facetClosing ? ' closing' : '')} data-field={title}>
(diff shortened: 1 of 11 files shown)


src/components/browse/components/FacetList/FacetOfFacets.js
```diff
@@ -0,0 +1,100 @@
...
+
+import { Collapse } from './../../../ui/Collapse';
+import { Fade } from './../../../ui/Fade';
+
+/**
+ * Used to render individual facet fields and their available terms in FacetList.
+ *
+ * @memberof module:facetlist
+ * @class Facet
+ * @type {Component}
```
alexkb0009: (Not Important Thoughts)
Is not really important to put too much stuff into docstring, lately have only been trying to add in docstrings for info that can't easily intuited from the code (e.g. usually not `@class`) but just to cut down on amount of lines of code sometimes.

At one point we had tried to use JSDoc, extract docstring, build reference documentation, and put it up on ReadTheDocs (docs website for Python libraries) which presented more work than was really worth. JSDoc required some verbose docstring, when working with ES5+ (classes, etc) especially. We then went to ESDoc, which was nicer, partly because it could intuit 'type', 'class' and other things which JSDoc couldn't without explicit docstrings. Then kind of stopped since it seems more useful to just have VS Code parse doc strings and give them when hover over an imported function/component or similar.

alexkb0009: If interested, the probably 2y+ old reference (JS stuff based on JSDoc) still persists at 4dn-dcic.github.io/fourfront/.
As can tell the JS docs aren't too useful.
We are able to build much nicer "ESDoc" reference, but were some (CSS/styling?) glitches when tried to upload to doc.esdoc.org and haven't had much time to look into further.

src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -9,6 +9,7 @@ import ReactTooltip from 'react-tooltip';
 import { Collapse } from './../../../ui/Collapse';
 import { Fade } from './../../../ui/Fade';
 import { PartialList } from './../../../ui/PartialList';
+import { anyTermsSelected, mergeTerms } from './index';
```
alexkb0009: Want to avoid importing components from index.js file(s), or from components/files that itself import component that are exporting.

Because then there's some circular dependencies are created - index imports FacetTermsList, FacetTermsList imports index, and things like tree shaking (removing unused/unimported functions from final compiled build.js file) might not work properly.

In most cases an error would be thrown when try to do that, and guessing is working now since mergeTerms isn't called until after constructor and such. Idk. 

----

Would be better to put these into their own file, e.g. `utility-functions.js` or something (and maybe move some other funcs to there as well) _or_ maybe to have them in FacetTermsList, and then in index.js, `import { FacetTermsList, anyTermsSelected, mergeTerms } from './FacetTermsList';`

src/components/browse/components/FacetList/index.js
```diff
@@ -285,7 +349,52 @@ export class FacetList extends React.PureComponent {
...
+                    groupedOnly.push({
+                        key: facet.grouping,
+                        title: facet.grouping,
+                        facets: [generateFacet(facet, i)],
+                        areTermsSelected
+                    });
+                    inGroupIndices[facet.grouping] = groupedOnly.length - 1;
+                }
+            }
+        }.bind(this));
```
alexkb0009: Prefer to use arrow func (`()=> { }`) if need to bind.
(Subjective on my part probably:) For anonymous functions especially; makes it easier when reading through code to understand where pure function vs a method or bound function is by presence of the `function` keyword, but wouldn't really work if also using `bind`.

src/components/browse/components/FacetList/FacetOfFacets.js
```diff
@@ -2,29 +2,69 @@
...
+        // Maybe we could pull it out into reusable controller component. Maybe. Very low priority.
         this.state = {
-            'facetOpen'     : typeof props.defaultFacetOpen === 'boolean' ? props.defaultFacetOpen : true,
+            'facetOpen'     : typeof props.defaultGroupOpen === 'boolean' ? props.defaultGroupOpen : true,
             'facetClosing'  : false,
             'expanded'      : false
         };
     }
 
+    componentDidUpdate(pastProps, pastState){
```
alexkb0009: Initially, `mounted` is always false (due to serverside render) and is used in `renderFacet` to help set `defaultFacetOpen`(and then `defaultGroupOpen`, based on that) which transitively are always false as well on initial render. We thus need this `componentDidUpdate` to pickup when FacetList has mounted.

alexkb0009: Most of this logic is exact same as in FacetTermsList, just using defaultGroupOpen instead.

src/components/browse/components/FacetList/index.js
```diff
@@ -337,64 +278,91 @@ export class FacetList extends React.PureComponent {
...
+                defaultFacetOpen = defaultFacetOpen || (!isStatic && _.any(filters || [], function(fltr){
+                    return (fltr.field === facetField + ".from") || (fltr.field === facetField + ".to");
+                })) || false;
+                return <RangeFacet {...commonProps} facet={facet} key={facetField} anyTermsSelected={anySelected} {...{ defaultFacetOpen, isStatic, grouping, fromVal, toVal }} />;
+            }
+            if (aggregation_type === "terms"){
+                const terms = mergeTerms(facet, filters); // Add in any terms specified in `filters` but not in `facet.terms` - in case someone hand-put that into URL.
+                const anySelected = anyTermsSelected(terms, facet, filters);
+                const isStatic = TermsFacet.isStatic(facet);
+                defaultFacetOpen = defaultFacetOpen || (!isStatic && _.any(filters || [], function(fltr){
```
alexkb0009: Maybe should remove the `!isStatic` check here. Even if static and selected, theoretically should be open. If is static then that prop might not be used (is 'always open'), but w.e...

src/components/browse/components/FacetList/FacetOfFacets.js
Bianca-Morris: I swear there's new pieces of 4DN and CGAP related documentation that I discover almost every day. Thanks!

src/components/browse/components/FacetList/FacetTermsList.js
Bianca-Morris: Ok, makes sense.

src/components/browse/components/FacetList/index.js
Bianca-Morris: Ah okay. I think it's just force of habit from writing ES5 at my old job. I'll try to remember to use those.

Bianca-Morris:
Looks good; I'm just going to run a quick test on my machine and then I'll start the merging process.

src/components/browse/components/FacetList/FacetOfFacets.js
alexkb0009: It's mostly onMount because windowHeight is used to calculate `maxTermsToShow` sometimes; a transition from closed->open then occurs onMount/didUpdate>mountChanged for those which should be open. Theoretically not too important & can hardcode `maxTermsToShow` value to maybe get simpler code + performance (no transition). (idk)

PR #13, opened 2019-11-17
Fix Search Bar/QuickInfoBar Filter Issues + New Search Within Results Feature

ugur-camoglu:
Pls handle this PR with https://github.com/4dn-dcic/fourfront/pull/1211

### DIFF
--- a/src/components/util/search-filters.js
+++ b/src/components/util/search-filters.js
@@ -599,16 +599,19 @@ export function convertExpSetFiltersTerms(expSetFilters, to = 'array'){
 
 
 /** Return URL without any queries or hash, ending at pathname. Add hardcoded stuff for /browse/ or /search/ endpoints. */
-function getBaseHref(currentHref = '/browse/', hrefPath = null){
+function getBaseHref(currentHref = '/browse/', hrefPath = null) {
     const urlParts = url.parse(currentHref, true);
-    if (!hrefPath){
+    if (!hrefPath) {
         hrefPath = urlParts.pathname;
     }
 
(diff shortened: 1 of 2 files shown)


src/components/util/search-filters.js
```diff
@@ -599,16 +599,19 @@ export function convertExpSetFiltersTerms(expSetFilters, to = 'array'){
...
-    if (!hrefPath){
+    if (!hrefPath) {
         hrefPath = urlParts.pathname;
     }
 
     const baseHref = (urlParts.protocol && urlParts.host) ? urlParts.protocol + '//' + urlParts.host + hrefPath : hrefPath;
     const hrefQuery = _.pick(urlParts.query, 'type', 'q');
-    if (hrefPath.indexOf('/search/') > -1){
-        if (typeof hrefQuery.type !== 'string'){
+    if (hrefPath.indexOf('/browse/') > -1) {
```
alexkb0009: Not a huge fan of ~ `if (hrefPath.indexOf('/browse/') > -1) { ...` because pathnames might differ a bit eventually between CGAP and 4DN, e.g. currently 4DN has a /browse/ page but CGAP does not.

Don't have any worthwhile suggestions/ideas right now and think is fine right now for merging. In future might be worth trying to think about how can be made more agnostic regarding `href` for future.

(OffTopic) Also the `hrefPath` param now seems a bit unnecessary since we parse `currentHref` anyways, but I might be forgetting some other reason for having it. Very low priority / for later but may be worth removing param and `if (!hrefPath){ .. }` wrapper.

utku-ozturk: @alexkb0009, can we merge both this PR and https://github.com/4dn-dcic/fourfront/pull/1211 into master?

PR #14, opened 2019-11-18
Fix of Enum Fields in Submission View Page

utku-ozturk:
**Problem:**

- If a property is defined as enum in schema
and
- its value type is not string

then submission view fails while validating the submitted enum value, since it is recognized as string. This problem occurred while submitting a page with redirection. Page.json's related property is:
```

"code" : {
     "title" : "Response Code",
     "description" : "Code returned by response.",
     "comment" : "See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection",
     "type" : "integer",
     "default" : 307,
     "enum" : [301, 302, 303, 307]
}
```

**Solution:**

The value is converted to the schema-defined data type before validation/submission. (currently supports integer, float, number and boolean.)

### DIFF
--- a/es/components/forms/components/submission-fields.js
+++ b/es/components/forms/components/submission-fields.js
@@ -322,8 +322,24 @@ function (_React$PureComponent) {
           nestedField = _this$props2.nestedField,
           fieldType = _this$props2.fieldType,
           linkType = _this$props2.linkType,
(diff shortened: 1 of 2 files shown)


utku-ozturk:
@alexkb0009  my bad, forgot to remove console.log. just fixed, thanks.

PR #19, opened 2019-12-27
Display Search Header Static Section: search-info-header.<Item type>

utku-ozturk:
Pls handle this PR w/ https://github.com/4dn-dcic/fourfront/pull/1228

### DIFF
--- a/src/components/browse/components/AboveSearchTablePanel.js
+++ b/src/components/browse/components/AboveSearchTablePanel.js
@@ -188,13 +188,12 @@ const SearchHeaderSection = React.memo(function (props) {
     const title = (propTitle ? <h4 className="text-300">{propTitle}</h4> : null);
     const content = (propContent ? <BasicStaticSectionBody {...{ content: propContent, placeholderReplacementFxn }} filetype={filetype || 'txt'} /> : null);
 
-    return (
-        content ?
-            <div className="row mt-1">
-                <div className="col-12 col-lg-9 pull-right">
-                    {title}
-                    {content}
-                </div>
+    return content ? (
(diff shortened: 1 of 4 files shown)


src/components/browse/components/AboveSearchTablePanel.js
```diff
@@ -2,180 +2,44 @@
...
 import _ from 'underscore';
 import { patchedConsoleInstance as console } from './../../util/patched-console';
-import { load } from './../../util/ajax';
-import { getAbstractTypeForType } from './../../util/schema-transforms';
+import { BasicStaticSectionBody } from './../../static-pages/BasicStaticSectionBody';
 
-/** TODO IMPROVE */
 
-class AboveSearchTablePanelStaticContentPane extends React.Component {
+export const AboveSearchTablePanel = React.memo(function AboveSearchTablePanel({ context, placeholderReplacementFxn }) {
```
alexkb0009: (Very small change / almost irrelevant / unsure) - 

Moved the `React.memo` call up one level to `AboveSearchTablePanel` instead of `SearchHeaderSection`.
Reasoning for this is AboveSearchTablePanel receives 2 props, which inherit from Object (and thus the variables are object references), and thus the memo comparison operation (propA !== pastPropA || propB !== pastPropB) should be a lot faster than the comparison in `SearchHeaderSection` because `SearchHeaderSection` compares strings (lists of characters) and presumably might take a bit longer. Idk 🤷‍♂.

utku-ozturk:
[search-info-header.<Item type>](https://trello.com/c/ad4wSBjT/39-search-info-headeritem-type)

PR #21, opened 2020-01-27
Saytajax with display titles... still so broken

Bianca-Morris:
~So so broken.~ Mostly fixed!

**Known Bugs**:

- [x] ~Updated items are appended to nav on dropdown selection... UNLESS there's something typed into the `searchasyoutype` filter input. https://gyazo.com/316c110809cee5ae8673184b0f1ccef4~
- [x] ~Advanced search still has the appending problem regardless of if there's something in the dropdown input or not.~
- [x] ~Display titles in KeyDisplay not saved properly;~ newly selected/created objects don't use their `display_title` in hierarchy or input button if you submit a new item and then return to the edit page immediately afterward. This is fixed on page refresh. (I wonder if this is related to indexing somehow...) _Won't fix: this _ is _ an indexing related problem; basically provides a placeholder_
- [x] ~After creating a new object, if you delete or try to replace it it does not disappear from the submission tree~
- [x] ~Once an item is selected, some random number appears in the filter input when you reopen the dropdown. Usually a 1 or 0; maybe an index or something?~
- [x] If you create a new object and then select the (newly indexed) item from the SAYT dropdown, it deletes the value from top level state
- [x] Re-selecting the same item twice makes the item disappear. ~Probably~ related to above. **Semi-Fixed**: _basically check to see if the current value of the field you're selecting is the same as the item you're trying to select, and if so just "updates" with the same item. Otherwise, deletes it -- so if you're editing the same field with the same value, it doesn't disappear. But if you're selecting a different field in an array with a value that's already present elsewhere in the array, deletes the most recent selection. Wish there was a way to flash or highlight briefly the item that was duplicated, so it's apparent that it's deleting for a reason..._ **See: https://gyazo.com/8419a96cf3ddb90bde20a3a3d2d90e7d** Note, this does not work for sub-embedded objects like family members on CGAP. 
- [x] Attempting to create a new experiment set of type CaptureC throws an error. https://gyazo.com/c12482d6636efc2dca39b25cc2c72e79 **Short term fix implemented; still seeking better long-term alternative**
- [x] Fix TypeError `pointer is undefined` on submit newly created linked object on CGAP ~(might be related to above bug)~
- [x] ~Separate out 4dn-specific style updates into fourfront branch~ not necessary; was able to successfully override in CGAP
- [x] Write some separate cgap-specific styles
- [x] ~Fix keyHierarchy headers for CGAP subembedded objects https://gyazo.com/faff011a85f81571733e5536cf4ac1eb (put members under member heading, phenotypic features under phenotypic features heading, etc)~ won't fix in this PR. this was broken before any of this SAYTAJAX stuff.


Bianca-Morris:
Not quite ready to merge. I did run into a couple of pretty big bugs:

**1)** This one managed to break every page that pulled data. White screen of death; seems like App.js could get maybe a quick flash of a render working and then everything broke. Related to Google Analytics.
Screen Shot 2020-01-30 at 5 22 22 PM
More stacktrace:  _https://gyazo.com/2e89112b0fdcac2c2bfab85764289f6d_


**2)** And then this one was actually, I think, the precursor to #1 above. Noticed this on a few of the pages I was looking at before first encountering the GA bug. Might not be related, but who knows.
Screen Shot 2020-01-30 at 5 24 36 PM

I'll be honest, I'm unsure exactly what caused either of these, and haven't been able to replicate them since encountering them the first time. Will keep testing and see if I can figure anything out.

src/components/forms/SubmissionView.js
```diff
@@ -630,33 +665,46 @@ export default class SubmissionView extends React.PureComponent{
             const linksCopy = _.clone(keyLinks);
             const roundTwoCopy = roundTwoKeys.slice();
             const hierarchy = _.clone(keyHierarchy);
-            let dummyHierarchy = object.deepClone(hierarchy);
-            let hierKey = key;
+            const dummyHierarchy = object.deepClone(hierarchy);
+
+            let keyToRemoveIdx = parseInt(keyToRemove) ? keyToRemove : null;
+            const keyToRemoveAtId = !parseInt(keyToRemove) ? keyToRemove : null;
```
alexkb0009: Not sure if properly checking what want it to -
I think more likely we want `!isNaN(keyToRemove)` (?)
isNaN returns true if not a number or false if it is a number, regardless of type (string or number).
ParseInt always return an integer, however that integer might be "NaN". ~~I believe "NaN" evaluates to truthy so `parseInt(keyToRemove)` would transitively always be `true`.~~

`NaN` evaluates to "falsy" so we're ok, however parseInt("0") would return 0 which is also falsy. So something to consider.

```javascript
>> parseInt("abcd")
NaN
>> !!(parseInt("abcd"))
false
>> !!(parseInt("123"))
true
>> !!(parseInt("0"))
false
>> isNaN("1")
false
>> isNaN("a")
true
```

src/components/forms/components/SearchAsYouTypeAjax.js
```diff
@@ -0,0 +1,699 @@
+import React, { useMemo } from 'react';
+import PropTypes from 'prop-types';
+import { _ } from 'underscore';
+import memoize from 'memoize-one';
+import ReactTooltip from 'react-tooltip';
+
+import { Fade } from './../../ui/Fade';
+
+import { ajax, object } from './../../util/';
+import { valueTransforms } from './../../util';
```
alexkb0009: These 2 lines can be combined into `import { ajax, object, valueTransforms } from './../../util/';`

src/components/forms/components/SearchAsYouTypeAjax.js
```diff
@@ -0,0 +1,699 @@
...
+
+import { Alerts } from './../../ui/Alerts';
+
+import { LinkToSelector } from './LinkToSelector';
+import { SearchSelectionMenu } from './SearchSelectionMenu';
+
+export class SearchAsYouTypeAjax extends React.PureComponent {
+
+    static getRegexQuery(value, filterMethod) {
```
alexkb0009: We should remove this method as well as `filterOptions`, at least from this component, since only relevant for SearchAsYouTypeLocal.

src/components/forms/components/SearchSelectionMenu.js
```diff
@@ -1,52 +1,110 @@
...
+        }
+    };
+
     constructor(props){
         super(props);
-        this.onToggleOpen = this.onToggleOpen.bind(this);
         this.state = {
-            dropOpen: false
+            dropOpen: false,
+            refreshKey: 0 // iterated to force a refresh of dropdown
```
alexkb0009: `// incremented ..` :-P

Is this (+ set in componentDidUpdate) ~~really necessary (since it seemed to auto-re-layout on scroll) or this handle other things?~~ bleh sorry I need to read before typing, realizing is re: options changing.

Not important but for fun might consider ways to make this slightly more performant if possible, i.e. if we have a max height that can only fit ~ 20 options, could do like if `Math.min(20, oldOpts.length) !== Math.min(20, newOpts.length)` to reduce # of state updates so that # of results change from like 97 to 340, and it doesnt visually impact dropdown size, can skip the increment. Idk probably not the added code since likely is already more than performant enough.

Another thought (though unsure if better) is can maybe call `this.forceUpdate()` to trigger a refresh of component without needing an additional state field (refreshKey).

3rd thought (probably wouldn't go this route mostly because we don't have control of React-Bootstrap and their API or whatever might change) is maybe if could pass down refs to React-Bootstrap's elements, access popper instance, and then call `popper.relayout()` or something or other. Since PopperJS isn't a React thing, theoretically for perf. gain should be fine to bypass React (aside from detecting componentDidUpdate) to manage it.

Bianca-Morris: Whoops on the incremented. 

The problem I have with max heights is that once we add in more "customized item selections" with data for the specific object (showing age, gender for individuals, for instance), that max height per dropdown item won't always be consistent, so we'll have to access one of the nodes directly. I'm not 100% sure, but I'm thinking that could end up being really complicated due to the way that `SearchSelectionMenu` is implemented. Since `Dropdown.Item`s are all rendered inside of a map function, even if we were able to assign refs to one/some of them, they'd end up being references to React components, and I'm not sure how to get the height of the rendered html element out of that... would we use findDomNode or something?

I was thinking about using forceUpdate, but for the reasons listed [here](https://stackoverflow.com/questions/30626030/can-you-force-a-react-component-to-rerender-without-calling-setstate) didn't do so. Also found [an issue](https://github.com/mui-org/material-ui/issues/13179) with almost the exact same problem, and they ended up going this way, as well. So it seems like this is at least a somewhat standard way of handling this.

Yeah, I considered that too when I saw that in Popper's docs, but I'm thinking that anything that requires going through React-Bootstrap to get to the Popper element is probably a no-go. React-Bootstrap doesn't seem to give us a clean-cut way to access the Popper instance through their API; this is something I ran into when I thought the issue was a Popper ref issue and was trying to find a way to check exactly what reference & popper elements were being passed into the Popper element.

So yeah, IDK. I'm not sure what the best way to limit the re-renders would be. In the meantime we can probably come back to it if it becomes an issue.

src/components/forms/components/SearchAsYouTypeAjax.js
Bianca-Morris: Really good catch; didn't even realize it wasn't being used anywhere.

PR #23, opened 2020-02-25
FacetList Updates re: UX -

alexkb0009:
Selected terms are now always visible, even if facet is collapsed per UX research / design / recommendations from Shannon.

- If all terms are selected, no option to collapse until deselect term(s).
- For simpler transitions, active (selected or omitted) terms are grouped at top of list now.

Also, minor fixes UI/styling edits esp re: new UX changes.
Most notably, for "Common Properties", no terms get moved down there unless no filter set for that field at all anywhere. This gives a more much predictable experience without selected filters (often unexpectedly) jumping down into "common properties".

Selected (and/or Omitted) terms now always show up the same - as highlighted bar(s) - even if facet containing the term is collapsed. Gives a large quick overview of currently-active filters.

### Screenshots

Expanded:

Collapsed:

### Future

May introduce a "collapse all facets" icon/button in header, next to clear all filters. Needs some code refactoring so will wait until get more feedback on these edits.


utku-ozturk:
We also tested on local and works great!

alexkb0009:
Thanks guys!

PR #24, opened 2020-03-01
HideColumns Prop Introduced to EmbeddedSearchView 

utku-ozturk:
`HideFacets` like new `HideColumns` prop introduced to hide columns if default columns will be displayed in `EmbeddedSearchView`.


utku-ozturk:
@alexkb0009 already implemented `listToObj` func caches facets' both `equals` and `not equals` values, and so memoizing makes sense. (+ not sure but hidden facets count mostly be greater than hidden columns) On the other hand, hideColumns have a few items as you wrote, memoize perf. gain/lose is insignificant.

PR #28, opened 2020-04-12
Search Results Table: Multiple Sort Options (per column); CSS+layout updates (more flexbox)

alexkb0009:
**Should be merged with CGAP PR https://github.com/dbmi-bgm/cgap-portal/pull/109 & 4DN portal updated re: new columnExtensionMap render function signature**

## Changes

This is mostly refactoring of various things relating to SearchResultTable in order to allow in CGAP to create table cells which have multiple lines, as well as to allow multiple sort options. Some code/functionality that is not too clean has been removed (i.e. forceUpdate method usage and such). Flexboxes allow lot more layout possibilities with ease than historical pre-CSS3 approaches.

Following this, "Display Title" column block styling has been updated a bit to take advantage of flexbox display. "Width" as a prop (of the table cell) becomes largely irrelevant for most things and might be worth stopping passing it down to ResultRowColumnBlock at all (aside from where is applied as width of element).

What was started on at one point was converting SearchResultTable cells from having `display: inline-block` or `float: left` (whichever it was back in the day) to `display: flex`, this branch largely attempts to finish that up. The actual component to render multiple values will be created in context of Variants Item data, once it exists.

## Previews

[Video Preview as of 04/12/2020](https://gyazo.com/af697255aef82def14a9d8326f413b00)

Screenshot as of 04/13/2020:

(+ "multiple options" indicator appended to bottom right of sort icon)

Screenshot as of 04/14/2020:

This PR is first of multiple ones for table improvements re: new CGAP UXD.

Should be relatively self-explanatory; UX should be what think should be heuristically expected for such feature and code to enable it is stable/readable/performant.

The following PR will be styling/content of results row re: new CGAP UX+design+styling (e.g. col striping).

_Low-Priority Longer Term Stuff:_
- Consider to no longer ever use React-Bootstrap and instead favor Bootstrap (official library) CSS + own JS. If possible. Lmk thoughts.
  - And ~~later~~ maybe way later to refactor/remove react-bootstrap dependency completely, creating drop-in replacements for still-used common components such as `DropdownButton` (straightforward: global singleton class tracks all dropdown instances to only allow 1 to be open at once, close when click off).
- Styling for CGAP dropdowns to match CGAP branding intents (any that we know of now, anyway) slightly better. Something better than plain white BG + 1px border + border radius.

### DIFF
--- a/es/components/browse/SearchView.js
+++ b/es/components/browse/SearchView.js
@@ -148,6 +148,7 @@ function (_React$PureComponent) {
       var childViewProps = _objectSpread({}, passProps, {
         currentAction: currentAction,
         schemas: schemas,
+        windowWidth: windowWidth,
         isOwnPage: true,
         facets: propFacets || contextFacets
       });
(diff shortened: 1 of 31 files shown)


alexkb0009:
Ready for review, may still push minor edits here and there only to improve interface/props/etc if needed for changes re: CGAP styling/design.

PR #29, opened 2020-04-22
fix for findValidationState; update parent valid. state on obj submission

Bianca-Morris:
I ran into a bug in SubmissionView state that was causing the final create new linked object test in my SAYTAJAX cypress tests to fail. Apparently, we weren't checking/updating the validation state of the parent item anywhere upon submission of a newly created item. (I still sort of can't believe this? Was this maybe implemented at some point previously, and we accidentally deleted it? Or do you happen to know of something like this happening later in the flow of the 'create new linked item'-related methods?) 

Anyway, this resulted in there being occasions where the parent item had no more incomplete child objects, but was still unable to be validated (validate button disabled). So basically, I added a check before the state is set to see if the parent item's validation status would be changed based on the submission of the current object, and under the circumstances in which it is, added that change to the state update.

But that didn't actually work because at some point during the creation of SAYTAJAX we had to change the way the keys for recently submitted items were stored in keyHierarchy (instead of using the temporary index, newly submitted items are now keyed with their @id). I don't remember _exactly_ we made this change; maybe you can remind me. But it had the side effect of making the findValidationState function basically useless, since the conditions it was using to determine whether or not an item was complete were no longer consistent with the rest of the code.

This PR aims to fix both of those problems, and should have validation working properly again when new items are created. Although, if you could, please also test some other cases, like adding via advanced search, etc. I feel like it's too easy to break things in this view, and adv search is one area where we severely lack test coverage, unfortunately. Thanks.

### DIFF
--- a/src/components/forms/SubmissionView.js
+++ b/src/components/forms/SubmissionView.js
@@ -362,7 +362,7 @@ export default class SubmissionView extends React.PureComponent{
      * lookup even if applicable and move right to alias selection.
      */
     initCreateObj(ambiguousType, ambiguousIdx, creatingLink, init=false, parentField=null){
-        console.log("calling initCreateObj with:", ...arguments);
+        // console.log("calling initCreateObj with:", ...arguments);
         const { schemas } = this.props;
         const itemTypeHierarchy = schemaTransforms.schemasToItemTypeHierarchy(schemas);
         // check to see if we have an ambiguous linkTo type.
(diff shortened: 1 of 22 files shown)


src/components/forms/SubmissionView.js
```diff
@@ -1298,9 +1297,15 @@ export default class SubmissionView extends React.PureComponent{
                                 this.setState(stateToSet);
                             }
                         } else {
+                            // Check if parent validation state will change based on current submission... update that alongside rest of state, if so
+                            const newParentValidState = SubmissionView.findValidationState(parentKey, stateToSet.keyHierarchy, stateToSet.keyContext, stateToSet.keyComplete);
+                            if (newParentValidState !== stateToSet.keyValid[parentKey]) {
```
alexkb0009, line 1302: Theoretically the 'if' check is not needed since if remains same, then.. nothing really differs upstream. PureComponent/React.memoized components will see the same value and not react to it.

PR #30, opened 2020-05-05
File Upload Drag and Drop

Bianca-Morris:
# Overview
This component allows users to press a button to open up a drag and drop window that will upload a new file and link it to the currently selected Item (assuming there's an appropriate field with attachments). Designed primarily for use with CGAP's pedigree sidebar tab, with a secondary objective of keeping the structure such that it could be eventually used in SubmissionView, too.

**Test with: [_'file-upload-dnd'_ branch on CGAP](https://github.com/dbmi-bgm/cgap-portal/pull/116)**

These are the main components in the SPC directory:

- `DragAndDropUploadFileUploadController`: The main controller. Handles uploading, linking, and state for dropped files. Most of this is probably self-explanatory. 
   - One thing to note: if multiple files are being uploaded in a single batch, the upload promise chain kicked off in `onUploadStart` is controlled by `PromiseQueue`. This forces each file's full promise chain (validate => create => link) to complete before starting to upload the next file, so that it can capture the new items' atIDs and make sure they are PATCH'd alongside any pre-existing items.
   - May rework the batching/queueing eventually; there were a bunch of features I'd originally intended to add that turned out to be too intensive to continue with that informed the decision to implement it this way, so it isn't maybe as optimal as it could be for the current implementation. If you have ideas on how to improve this, LMK.
- `DragAndDropUploadButton` and `DragAndDropModal`: The button, when clicked, spawns the modal. The modal updates according to upload state in the Controller. 
    - Originally intended to also use a file input for browsing on click, as well, but that's a bit more complicated than I expected. (https://medium.com/trabe/controlled-file-input-components-in-react-3f0d42f901b8) If it's a desired feature, happy to add, but since it wasn't explicitly requested, decided not to include in this initial version.
- `DragAndDropZone`: Handles all of the actual event handling for the dropzone itself.

# Known Bugs:
- If you change window width from non-tablet to tablet, the modal is forced closed, along with the sidebar. Think this was intended functionality for pedigree, but may have implications for uploads in-process. Not likely to matter in MVP use-case, so didn't fix.
- If there are many (>10 or so) files, the pedigree sidebar pane will expand slightly, making the pedigree background appear to stop early. Working on a quick fix for this.
- Never figured out how to get the range for allowed file size from schema (is there somewhere else I should be looking for that?) so didn't add that. So, the error you'll get on too large a file upload will be from the server, not client side.


src/components/forms/components/DragAndDropUpload.js
```diff
@@ -0,0 +1,674 @@
...
+
+/**
+ * A class utility for executing promise-chains in a sequential manner. Includes some
+ * scaffolding for aborting promises; needs more work in future.
+ *
+ * In drag-and-drop upload, this class ensures that each Item is uploaded and linked
+ * before starting the upload of the next item, so that the new atIds can be collected
+ * and patched to the parent together.
+ */
+class PromiseQueue {
```
alexkb0009, line 20: All the properties and methods of this class are static, meaning there is ever one single value/method shared between all instances of the class (technically is case with instance methods, given prototypical inheritance in JS objects/classes, but for instance methods, those methods are bound to `this` of the outer class/function scope.)

This is ok if want the class to be a singleton and only ever have one instance of it.. but I think we actually _might_ theoretically have multiple drag-drop areas or similar, so think would be better to have these as instance methods/properties (?).

src/components/forms/components/DragAndDropUpload.js
```diff
@@ -0,0 +1,674 @@
...
+    generatePayload(file, attachmentPresent) {
+        const { award, lab, institution, project } = this.props;
+
+        const aliasFilename = file.download.split(' ').join('');
+        let alias;
+
+        const payloadObj = {};
+
+        // If on 4DN, use lab and award data (institution/project should be null)
+        if (lab && award) {
```
alexkb0009, line 214: (For future) Great for now; we should remove this logic here and in SubmissionView (and have it be done for us in backend). Gotta wait for backend to do this first though I think / we could check in w. Kent & Will later.

src/components/forms/components/DragAndDropUpload.js
Bianca-Morris: Hmm this is a good point. I was thinking, I think, that if someone started adding for Documents and then, while those things were being uploaded also started uploading some Images that they might be able to all end up in the same queue. I'm honestly not sure if that's how it would work, though, as I haven't been able to upload something big enough/that takes long enough to test it.

Bianca-Morris: Ok updated this to use instance-specific methods. And I create a new instance of it when it is used inside of the `DragAndDropFileUploadController.onUploadStart` method. So each separate Drag and Drop thing should now be using its own queue.

alexkb0009:
Might put PromiseQueue into own file and repurpose for other things later (i.e. authentication or similar).

PR #31, opened 2020-05-25
Render linkTo Objects[] as link elements in table cells

alexkb0009:
If have array of objects (linkTos) as value, then output array of link elements instead of ~ [Object object]. Should be self-explanatory from code. Output could look like, if change 4DN ExperimentSetTables.js defaultProps.columns `"experiments_in_set.biosample.biosource.individual.organism.name"` to `"experiments_in_set.biosample.biosource.individual.organism"`:

### DIFF
--- a/es/components/browse/components/table-commons/ResultRowColumnBlockValue.js
+++ b/es/components/browse/components/table-commons/ResultRowColumnBlockValue.js
@@ -129,9 +129,28 @@ function (_React$Component) {
         return null;
       }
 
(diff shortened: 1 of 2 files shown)


utku-ozturk:
I just checked w/ my current dataset (includes labs and enzymes but not experiment-hi-c) and it works great. Code looks also great but not sure about className comparison `value[0].props.className === "link-wrapper"`. In any code change, it may be prone to error. Moreover, your approach is robust compared to the branch merged.

PR #32, opened 2020-05-28
QC Summary Schema Update

utku-ozturk:
https://github.com/4dn-dcic/fourfront/pull/1326


alexkb0009:
I didn't test locally but looks good to merge!
After merge, would create a new SPC release with new minor version number, update to that release in fourfront branch, and then probably merge fourfront branch (will look at it next).

Also checkout [JavaScript Destructuring](https://hacks.mozilla.org/2015/05/es6-in-depth-destructuring/) (probably better/other articles on this exist). I found it helps keep the amount of code/text down (and more readable) when a property **may** or **may not** be present, especially. (This might be subjective-ish on my part, lmk what think).

utku-ozturk:
@alexkb0009 thank you for your valuable comment. Honestly, I was unaware of nested destructuring in detail and read the link you referenced and other articles about destructuring at weekend. I refactored the code and it is more readable for me and of course for everyone.

PR #33, opened 2020-06-23
Handle date facets, minor fix

alexkb0009:
Should work as expected, looks like the following:

Uses (most-) browser's [built-in type="date" input elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date), which often provide a calendar select widget.
In older/unsupported browsers, this should degrade to a plain text input.

#### Future Considerations

1. Fallback for unsupported browsers
2. Do we ever want increments of dates ? (current assumption is no)
3. Redirect from `field.from=2018` to `field.from=2018-01-01` perhaps...

### DIFF
--- a/src/components/ui/LocalizedTime.js
+++ b/src/components/ui/LocalizedTime.js
@@ -2,38 +2,51 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import memoize from 'memoize-one';
 import moment from 'moment';
 import { isServerSide } from './../util/misc';
 
 
-export class LocalizedTime extends React.PureComponent {
+export class LocalizedTime extends React.Component {
 
     constructor(props){
         super(props);
-        this.state = {
-            moment : props.momentDate ? props.momentDate :
-                props.timestamp ? moment.utc(props.timestamp) : moment.utc(),
-            mounted : false
+        this.memoized = {
+            getMoment: memoize(function(momentDate, timestamp) {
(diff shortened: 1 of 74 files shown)


src/components/ui/LocalizedTime.js
```diff
@@ -2,38 +2,51 @@
...
-export class LocalizedTime extends React.PureComponent {
+export class LocalizedTime extends React.Component {
 
     constructor(props){
         super(props);
-        this.state = {
-            moment : props.momentDate ? props.momentDate :
-                props.timestamp ? moment.utc(props.timestamp) : moment.utc(),
-            mounted : false
+        this.memoized = {
```
alexkb0009, line 14: This returns new moment instance if a new timestamp occurs. Previously, if timestamp changed, this wouldn't update in response.

src/components/util/analytics.js
```diff
@@ -29,7 +29,7 @@ const defaultOptions = {
             track_and_facet_info: { experiment_type: tfi_expType } = {},
             // For exps or expsets, try to grab experiment_type
             experiment_type: { display_title: exp_expType } = {},
-            experiments_in_set: [ { experiment_type: { display_title: set_expType } = {} } ] = [{}],
+            experiments_in_set: [ { experiment_type: { display_title: set_expType } = {} } = {} ] = [{}],
```
alexkb0009, line 32: Fixes a glitch (hopefully / needs testing)

PR #38, opened 2020-08-12
External References Updates

utku-ozturk:
**First merge this PR, then release a new version. Then update SPC version of https://github.com/4dn-dcic/fourfront/pull/1367 and merge.**


utku-ozturk:
ItemFooterRow.js has been moved from SPC to FF since there is not any plans to use it in CGAP at the moment. Then ExternalReferencesStackedTable has also been moved into ItemFooterRow to group all related stuff.

PR #43, opened 2020-09-24
Extended Description Popover & Some SCSS edits

alexkb0009:
#### Primary Edit:

- Handling "extended_description" which may be present on schemas as a property of schema field or facet field definition.
- Overlay+Popover should be displayed for this.
- Related: https://github.com/dbmi-bgm/cgap-portal/pull/211

screenshot

#### Secondary Edits:
Many of the following edits are now possible due to 4DN and CGAP both being on Bootstrap V4. For 4DN, we should also update Bootstrap and React-Bootstrap to latest versions, but it isn't urgent.

- Renamed inline-block to d-inline-block so that we may remove the .inline-block class out of CGAP and 4DN repos, since class w. same purpose (d-inline-block) already exists in BootstrapV4.
- Similarly, removing .text-ellipsis-container in favor of .text-truncate which exists already in Bootstrap V4.
- Updated schemaTransforms to not return contents of "items" schema property (parent code can get that if needed easily).

#### Merging (and subsequent changes needed on 4DN & CGAP)

It should be safe to merge this and update both 4DN and CGAP versions to it. There are some 'breaking changes' in terms of us needing to pass `addToBodyClassList` & `removeFromBodyClassList` into EmbeddedItemSearchView from 4DN and/or CGAP respectively, but not an issue if no "extended_descriptions" yet exist in schemas. And if were, simply "stop body from scrolling" part of this feature wouldn't work, while other things remain functional.


alexkb0009:
Still finishing up some edits for this from CGAP-side (namely re: typography), but this is ready for review. There will be some edits needed on 4DN-side but should be minor.

src/components/browse/components/FacetList/ExtendedDescriptionPopoverIcon.js
```diff
@@ -0,0 +1,101 @@
...
+        <React.Fragment>
+            { adjustedChildren }
+
+        </React.Fragment>
+    );
+}
+
+
+/**
+ * @todo Potentially refactor to use ReactDOM.createPortal along w. semi-opaque dark background that will close this onClick.
```
Bianca-Morris, line 31: Random question: is ReactDOM.createPortal actually in use anywhere on CGAP currently? I remember I was looking around the repo for an example a few weeks back when I was messing around with creating an interpretation/annotation overlay. It looked like there was a lot of infrastructure to support the use of it, but that it wasn't actually being used yet.

src/components/browse/components/FacetList/ExtendedDescriptionPopoverIcon.js
```diff
@@ -0,0 +1,101 @@
...
+        title: facetTitle,
+        field,
+        extended_description: facetExtendedDescription = null
+    } = facet;
+    const {
+        title: fieldTitle,
+        extended_description: fieldExtendedDescription = null
+    } = fieldSchema || {}; // May be null for 'calculated' or 'fake' fields.
+
+    //const [ isOpen, setOpen ] = useState(false);
```
Bianca-Morris, line 46: Can probably get rid of this now that some of those variables are in use elsewhere.

src/components/browse/components/FacetList/index.js
```diff
@@ -562,54 +598,67 @@ export class FacetList extends React.PureComponent {
...
+                ) : null }
+            </React.Fragment>
         );
     }
 }
+
+// TODO: Pull out the split terms into own component
+// 2: get ourselves the fieldSchema and pass it down in here.
+
+// function
```
Bianca-Morris, line 663: I'm not clear on what this means or if it's left over/unnecessary? If it is important, might be good to clean it up a bit.

Bianca-Morris (approved):

src/components/browse/components/FacetList/ExtendedDescriptionPopoverIcon.js
alexkb0009: It's being used for [BigDropdownMenu and the 'actions' slide out pane on ItemViews](https://github.com/dbmi-bgm/cgap-portal/search?q=ReactDOM.createPortal) currently.

src/components/browse/components/FacetList/index.js
alexkb0009: Yeah is some leftover comments. Had noticed that TermsFacet doesn't do a whole lot and thought about combining the files TermFacet, FacetTermsList, and StaticSingleTerm. But on second thought unsure is great idea to make 1 big(ger) file; if anything could create a subfolder 'TermsFacet' and put those other things in it. But not really important enough to do now; maybe if start to add more functionalities to FacetList in near future, idk.

PR #44, opened 2020-10-14
SearchView & Related: Enabling aboveTableComponent etc.

alexkb0009:
### Changes

- Remove support for `props.topLeftChildren` & `props.selectedFiles`
- Enable support for `props.aboveTableComponent`
- Enable support for `props.aboveFacetListComponent`
- Enable support for `props.detailPane`
- (Unlikely) Might consider removing `props.renderDetailPane` in future sometime in favor of just `detailPane`


utku-ozturk (approved):
We have tested both default and customized (File/Publication/MicroscopeConf) search views locally, works well.

PR #46, opened 2020-10-21
getNestedProperty Func Updates

utku-ozturk:
Re: Trello ticket (https://trello.com/c/DWxkTepi)

The column value was not rendered if the nested property is a child of an array field and not exists for all array items.

E.g. https://data.4dnucleome.org/static-sections/5c02ae63-564e-4839-b8b4-ad8bb6939863/ has a `Total Reads` column to display `processed_files.quality_metric.Total reads` nested property. The `processed_files` field is an array and in this case some processed files have a valid quality_metric field but some have not. The current implementation totally ignores the field if all of them are not valid. We updated it to display valid nested property values if exists.

**Current:**

Screen Shot 2020-10-22 at 02 25 37

**Updated:** 

Screen Shot 2020-10-22 at 01 47 08

### DIFF
--- a/src/components/util/object.js
+++ b/src/components/util/object.js
@@ -158,8 +158,16 @@ export function getNestedProperty(object, propertyName, suppressNotFoundError = 
 
             if (Array.isArray(currentNode)){
                 var arrayVals = [];
-                for (var i = 0; i < currentNode.length; i++){
-                    arrayVals.push( findNestedValue(currentNode[i], fieldHierarchyLevels, level) );
+                for (var i = 0; i < currentNode.length; i++) {
+                    if (typeof currentNode[i][fieldHierarchyLevels[level]] !== 'undefined') {
+                        arrayVals.push(findNestedValue(currentNode[i], fieldHierarchyLevels, level));
+                    } else {
+                        arrayVals.push(null);
+                    }
+                }
+                if (!_.any(arrayVals, (val) => val !== null)) {
+                    if (!suppressNotFoundError) throw new Error('Field ' + _.clone(fieldHierarchyLevels).splice(0, level + 1).join('.') + ' not found on object.');
+                    return null;
                 }
                 return arrayVals;
             } else {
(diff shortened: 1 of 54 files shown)


src/components/util/object.js
```diff
@@ -158,12 +158,20 @@ export function getNestedProperty(object, propertyName, suppressNotFoundError =
...
-                for (var i = 0; i < currentNode.length; i++){
-                    arrayVals.push( findNestedValue(currentNode[i], fieldHierarchyLevels, level) );
+                for (var i = 0; i < currentNode.length; i++) {
+                    if (typeof currentNode[i][fieldHierarchyLevels[level]] !== 'undefined') {
+                        arrayVals.push(findNestedValue(currentNode[i], fieldHierarchyLevels, level));
+                    } else {
+                        arrayVals.push(null);
+                    }
+                }
+                if (!_.any(arrayVals, (val) => val !== null)) {
```
alexkb0009, line 168: Might be more clear to write `if (_.any(arrayVals, function(val){ return val === null; }))` (avoiding double-negative '!').

----

This function is meant to be used in many places (not just search result columns) so would double check that other places that use it aren't negatively affected (shouldn't be, though). One potential consideration is if array full of `null`s or mixed values including `null`s is an actual value we want to get -- which if there is such value, we probably _do_ want to grab it instead of receiving a single `null` or error. It might be better to filter such array outside the scope of this function if so.

utku-ozturk: @alexkb0009 `if(!_.any(arrayVals, (val) => val !== null))` might be rewritten as `if(_.all(arrayVals, (val) => val === null))` instead of `if(_.any(arrayVals, function(val){ return val === null; }))`, since we want to ensure at least one item in array is not null to avoid `Error`.

utku-ozturk: Re: One potential consideration is if array full of nulls or mixed values including nulls is an actual value we want to get

FYI, current implementation ignores the array altogether even only single item of the array is null.

alexkb0009 (approved):
Looks good!

PR #48, opened 2020-11-09
Range facet ui 1

Bianca-Morris:
Test with https://github.com/dbmi-bgm/cgap-portal/pull/265 (CGAP) and https://github.com/4dn-dcic/fourfront/pull/1422 (4DN). Still trying to work out some differences between 4DN and CGAP styling, and doing general cleanup (especially comments, no longer necessary code, etc) but basically ready.

### Upcoming To-Dos:

- [x] Update `FacetList.js` with an `onFilterMultipleExtended` method that processes analytics events for multi-filter updates
- [x] ~Change selection color + add tooltip for invalid selection~ removed
- [ ] ~Add "reset/X" button to dropdown for resetting back to original value now that we're auto-updating onBlur~ Re-add 'ESC' event handler for reset
- [x] Fix nested RangeFacet "0" value selection bug
- [x] More tweaks for rounding large numbers/decimals
- [x] Consider cost-benefit of removing one-click-clear
- [x] Make selection dropdowns disappear when facet is closed
- [x] ~Fix loading status indicator bug~ ended up getting rid of this entirely; should be easy to re-add if necessary
- [ ] Optimize filter block componentDidUpdate fix
- [x] Add indicator for currently selected preset range selections
- [ ] Smooth out collapse animation (optional)
- [x] Tweaks to ordering (optional)
- [ ] Improve/tweaks to styling (get rangedrop to fit at all button widths -- figure out why 4DN dropdown sometimes is overlapped)

**More testing, deploy for testing**

### Eventually: 
- [x] ~Discuss possibility of replacing onFilter with onFilterMultiple everywhere -- is it worth the extra work?~ _Nope_


src/components/browse/components/FacetList/RangeFacet.js
```diff
@@ -411,7 +411,7 @@ export class RangeFacet extends React.PureComponent {
             <div className={"facet range-facet" + (facetOpen ? ' open' : ' closed')} data-field={facet.field}>
                 <h5 className="facet-title" onClick={this.handleOpenToggleClick}>
                     <span className="expand-toggle col-auto px-0">
-                        <i className={"icon icon-fw icon-" + (isOpen ? "minus fas" : "plus fas")}/>
+                        <i className={"icon icon-fw icon-" + (facetOpen ? "minus fas" : "plus fas")}/>
```
alexkb0009, line 414: If had a value, but facet is closed, this icon wouldn't change (remain minus), so this addresses that. We could consider renaming variable 'isOpen' 🤔 lol.

alexkb0009 (approved):
Code looks good, functionality works great!

Would consider iterating on CSS/styling a tiny more if time etc allows (e.g. maybe increase btn widths to fit closer to half of available width, unless small window width, then maybe stack (?) (or not idk)).

PR #52, opened 2020-12-11
SubmissionView SAYTAjax Dropdown Closes on First Click

utku-ozturk:
Trello ticket: https://trello.com/c/R4VOvNvX

**Solution:** a very minor update that removes redundant setState call in `componentDidUpdate` of `SearchSelectionMenu`

Updated_version

### DIFF
--- a/es/components/forms/components/SearchSelectionMenu.js
+++ b/es/components/forms/components/SearchSelectionMenu.js
@@ -59,7 +59,10 @@ export var SearchSelectionMenu = /*#__PURE__*/function (_React$PureComponent) {
           newOptions = _this$props$options === void 0 ? [] : _this$props$options;
       var refreshKey = this.state.refreshKey;
 
-      if (oldOptions.length !== 0 & oldOptions.length !== newOptions.length) {
+      if (oldOptions.length !== 0 && oldOptions.length !== newOptions.length) {
+        // TODO: calling setState in componentDidUpdate toggles SAYTAjax dropdown as closed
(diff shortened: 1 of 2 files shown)


Bianca-Morris (approved):
Ah, fascinating; such a simple fix. Thanks for taking this on. I know it has been a frustration for quite a while now.

alexkb0009:
If works looks good, re: the twitter link/post also reminds of another probably unrelated thing (+ @Bianca-Morris ) -- 

Best practice is to use a _function_ as body of a setState call esp if state update is derived from a previous/existing state -- this is because [state updates may be run asynchronously and/or merged together](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous).

Ideally should be something like ~
```javascript
this.setState(function(existingState, existingProps){
  const { refreshKey } = existingState;
  return { "refreshKey": refreshKey + 1 };
})
```

Have no idea if this will help in this PR/issue or not at all, but this is perfect analog to the example shown in the linked React state-and-lifecycle documentation --

utku-ozturk:
@alexkb0009 I have replaced setState statement w/ the code below but it does not help when `oldOptions.length !== 0 &&` condition is removed. If the code enters into if condition in the future, it will be more robust, so I merged it as is. 

``` js
this.setState(function(existingState, existingProps){
  const { refreshKey } = existingState;
  // todo: maybe read existingProps here as well and then cancel-out (return null) if no update needed.
  return { "refreshKey": refreshKey + 1 };
})
```

PR #53, opened 2020-12-17
Schema maxItem Edits

utku-ozturk:
Trello: https://trello.com/c/1rJiBc8a

If a schema has "maxItems": n, no more than n entries should be possible via UI.
In the example, "maxItems": 2, but the option to add a third Relationship Type appears after the second. Ideally, this should not happen.

Solution: submission-fields ArrayField is updated to check `maxItems`. If the total items count is equal to maxItems then we prevent adding a new item at the bottom.

https://i.gyazo.com/ca52b5a2ef67fe79e3584f143bbf85b3.gif

https://i.gyazo.com/28abcd19bed84c129bbef413c70169d8.gif


Bianca-Morris (approved):
Looks good.

I did notice a recurring error message while testing this though:
```
SubmissionTree.js:204 Uncaught TypeError: Cannot read property 'sort' of undefined
    at SubmissionLeaf.generateAllPlaceholders (SubmissionTree.js:204)
    at SubmissionLeaf.render (SubmissionTree.js:256)
    at finishClassComponent (react-dom.development.js:17160)
    at updateClassComponent (react-dom.development.js:17110)
    at beginWork (react-dom.development.js:18620)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
    at beginWork$1 (react-dom.development.js:23203)
    at performUnitOfWork (react-dom.development.js:22157)
generateAllPlaceholders @ SubmissionTree.js:204
render @ SubmissionTree.js:256
finishClassComponent @ react-dom.development.js:17160
updateClassComponent @ react-dom.development.js:17110
beginWork @ react-dom.development.js:18620
```
Probably not related to this PR (I know second-round submission has been buggy for a while now), but just letting you know in case you have a quick fix. Steps to reproduce: [here](https://gyazo.com/c7ee2ec30e0a05780738ffd589ead88d).

src/components/forms/components/submission-fields.js
```diff
@@ -489,12 +490,24 @@ class ArrayField extends React.Component{
 
     componentDidUpdate(prevProps, prevState){ // We can't do a comparison of props.value here because parent property mutates yet stays part of same obj.
         const { value, field, pushArrayValue, modifyNewContext, nestedField, schema, linkType } = this.props;
+        const { maxItems } = schema;
+
         if (ArrayField.shouldPushArrayValue(value, field)){
+            if (maxItems && value.length == maxItems)
```
alexkb0009, line 496: It is considered bad practice to use `==` instead of `===` in JavaScript. This is because `==` in JavaScript doesn't check type and things like `false == 0` and `false == ""` will evaluate to true, and this is often a source of hard-to-trace bugs. [More Info.](https://bonsaiden.github.io/JavaScript-Garden/#types.equality)

If know for a fact that are comparing two variables of the same type, then `==` is safe to use, but there's no real reason to since `===` is about as performant and also type-safe.

src/components/forms/components/submission-fields.js
```diff
@@ -552,7 +565,7 @@ class ArrayField extends React.Component{
         const schema = propSchema.items || {};
         const values = propValue || [];
         const valuesToRender = _.map( values.length === 0 ? [null] : values , function(v,i){ return [v, schema, i]; });
-        const showAddButton = !isValueNull(values[valuesToRender.length - 1]);
+        const showAddButton = !propSchema.maxItems && !isValueNull(values[valuesToRender.length - 1]);
```
alexkb0009, line 568: Similarly to previous comment, would consider if maxItems can ever be validly 0, because if so, then `!0` will evaluate to `true`. Alternatively could test `typeof propSchema.maxItems === "number"` to determine presence (or lack) of the propSchema.maxItems even if it is 0. In this case it makes sense that maxItems can never be below 1 so would say is fine to leave as is.

alexkb0009 (approved):
Looks pretty good, made a minor comment or two that worth checking out / making minor adjustments in response to; haven't tested on local but assume if was working fine on hotseat or your local then is good to go.

PR #55, opened 2021-01-12
Search in Facet Terms

utku-ozturk:
### Problem: 
[Trello ticket](https://trello.com/c/tmkAyEL4)

### Solution:
In order to search in facet terms, we added 2 new fields into schemas/facets section: `search_type` and `sayt_item_type`.

example schema facets section:

```json
"facets": {
  "variant.genes.genes_ensg.display_title": {
      "title": "Gene",
      "order": 1,
      "search_type": "sayt_without_terms", //possible values are: none|basic|sayt|sayt_without_terms
      "sayt_item_type": "Gene" //a valid item type (valid if search_type is either sayt or sayt_without_terms
  }
}
```

**search_type:**

- **basic:** displays a basic search input and filters only in returned values
- **sayt:** displays a SAYT component along with returned terms. SAYT lists items having type defined in `sayt_item_type` 
- **sayt_without_terms:** displays a SAYT component but returned values are not rendered. Only terms selected via SAYT are listed.
- **none:** neither basic nor SAYT search is rendered

**sayt_item_type:**

Taken into consideration if the search_type is `sayt` or `sayt_without_terms`. If it is not defined, then the type's default value is `Item`.

### screencasts:
basic: [this](https://trello-attachments.s3.amazonaws.com/5d8f8b1b9c43fb39fe46a9c6/5f931588d290cb888d1a7e31/687c8e59b5fc3999462b3bd7d16f3492/2384090548f8e465dad0360316200c0c.gif) and [this](https://trello-attachments.s3.amazonaws.com/5d8f8b1b9c43fb39fe46a9c6/5f931588d290cb888d1a7e31/530e2c32253a4e49d784a211e5b66348/850adfb9a44eb5ce0586ad3e8ec31f53.gif)
sayt: [this](https://i.gyazo.com/785bc13971eda147bd3fe2e38d1e8ad8.gif)
sayt_without_terms: [this](https://i.gyazo.com/ad2e7f74cb68622436f4a9d8175604a8.gif)

### Notes: 

https://github.com/dbmi-bgm/cgap-portal/compare/sayt_in_facets has an example schema/facets editing to enable search in facet terms.


src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -370,14 +420,34 @@ const ListOfTerms = React.memo(function ListOfTerms(props){
...
+            facetSearch = (
+                <div style={{ 'padding': '10px', 'fontSize': '0.875rem' }}>
+                    <input className="form-control" autoComplete="off" type="search" placeholder="Search"
+                        name="q" onChange={onBasicTermSearch} key="facet-search-input" />
+                </div>);
+        } else if ((searchType === 'sayt') || (searchType === 'sayt_without_terms')) {
+            const itemType = facet.sayt_item_type && typeof facet.sayt_item_type === 'string' && facet.sayt_item_type !== '' ? facet.sayt_item_type : 'Item';
+            const baseHref = "/search/?type=" + itemType;
+            facetSearch = (
+                <div className="d-flex flex-wrap" style={{ 'padding': '10px', 'fontSize': '0.875rem' }}>
```
alexkb0009, line 436: Use className "p-2" (or similar, depending on padding desired / what looks good) and className "small" or "text-small" or "text-smaller" instead of defining styles inline.

Inline styles, especially in JSX/React -- [where performance is negatively impacted by it generally](https://reactjs.org/docs/faq-styling.html#are-inline-styles-bad) -- but also in general/html, are discouraged whenever they can be avoided.

src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -370,14 +420,34 @@ const ListOfTerms = React.memo(function ListOfTerms(props){
...
+        // or show SAYT control if item search is available
+        let facetSearch = null;
+        if (searchType === 'basic') {
+            facetSearch = (
+                <div style={{ 'padding': '10px', 'fontSize': '0.875rem' }}>
+                    <input className="form-control" autoComplete="off" type="search" placeholder="Search"
+                        name="q" onChange={onBasicTermSearch} key="facet-search-input" />
+                </div>);
+        } else if ((searchType === 'sayt') || (searchType === 'sayt_without_terms')) {
+            const itemType = facet.sayt_item_type && typeof facet.sayt_item_type === 'string' && facet.sayt_item_type !== '' ? facet.sayt_item_type : 'Item';
```
alexkb0009, line 433: Try to use [destructuring](https://itnext.io/how-to-use-object-destructuring-in-modern-javascript-59758ebfb778), will result in shorter and more readable lines of code (albeit sometimes more lines of code)

alexkb0009:
Code looks pretty good (made some comments re: code style, but minor overall). However is a [little bit broken in 1 or 2 tested cases](https://gyazo.com/441b13c543153cea0bbf76577efe6ba0) -- thinking might be some `if` condition that isn't getting matched maybe -- but unsure yet / looking through code further.

alexkb0009:
For future - showing counts of results per term - 

- We should have the terms available for top 100 items, under 'facet' object.
- Can create an object/dictionary of `{ [termName]: count }` from these.
- Can then pass this dictionary down to a custom new Component and/or function that would be used to render list items (results of search as you type) in the 'dropdown menu' via being passed into [SearchAsYouTypeAjax's `props.optionRenderFunction`](https://github.com/4dn-dcic/shared-portal-components/blob/master/src/components/forms/components/SearchAsYouTypeAjax.js#L199-L207) and in this component, lookup the count (if any), and render it at right.

src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -184,6 +209,16 @@ export class Term extends React.PureComponent {
                 </a>
             </li>
         );
+
+        if (searchType === 'basic') {
+            // if (!searchText) { return facetListItem; }
+            const termAlreadyFiltered = _.any(filteredTerms, function (item) { return item.key === term.key; });
```
alexkb0009, line 215: This seems quite imperformant -- think about how to refactor...

There might be 100+ terms on any view (most likely not all would be SAYT terms/facets, but still many might be). If each of these 100+ terms then iterates over other filtered terms (is this same size as all terms, potentially?) then we get performance of like O(n^2) and with enough Terms, UI might start to hang upon changes/as-enter-text.

Would at least have another (memoized, probably) function for this check or have `filteredTerms` be an object for faster lookups (rather than iterating over each term in filteredTerms using `_.any`).

src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -401,7 +401,24 @@ const ListOfTerms = React.memo(function ListOfTerms(props){
...
+    // or show SAYT control if item search is available
+    let facetSearch = null;
+    if (searchType === 'basic') {
+        facetSearch = (
+            <div className="text-small" style={{ 'padding': '10px' }}>
+                <input className="form-control" autoComplete="off" type="search" placeholder="Search"
+                    name="q" onChange={onBasicTermSearch} key="facet-search-input" />
+            </div>);
+    } else if ((searchType === 'sayt') || (searchType === 'sayt_without_terms')) {
+        const itemType = facet.sayt_item_type && typeof facet.sayt_item_type === 'string' && facet.sayt_item_type !== '' ? facet.sayt_item_type : 'Item';
```
alexkb0009, line 414: https://www.smashingmagazine.com/2019/09/reintroduction-destructuring-assignment/

src/components/browse/components/FacetList/FacetTermsList.js
Bianca-Morris: Agree on performance note.

In reference to that edit, I think this is a bigger question that would need discussion at a UI meeting. I completely get what you're saying - most of the lists of items we'd be using for local search are small enough that SAYT isn't super necessary, but I think there might be a couple of cases where it really would be preferable (long lists of country names, for instance). So it's sort of a balancing act; not necessary for everything, but probably better to have for others.

src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -351,8 +402,26 @@ const ListOfTerms = React.memo(function ListOfTerms(props){
...
+                <input className="form-control" autoComplete="off" type="search" placeholder="Search"
+                    name="q" onChange={onBasicTermSearch} key="facet-search-input" />
+            </div>);
+    } else if ((searchType === 'sayt') || (searchType === 'sayt_without_terms')) {
+        let { sayt_item_type: itemType = '' } = facet || {};
+        itemType = typeof itemType === 'string' && (itemType.length > 0) ? itemType : 'Item';
+        const baseHref = "/search/?type=" + itemType;
+        facetSearch = (
+            <div className="d-flex flex-wrap text-small p-2">
+                <SearchAsYouTypeAjax baseHref={baseHref} showTips={true} onChange={onSaytTermSearch} key={itemType} />
```
Bianca-Morris, line 420: Some styling notes for the dropdown: 
- I think the button would look nicer if it were bootstrap size "sm" rather than standard. Strikes me as a little large, currently. I don't remember if SAYTAJAX has a functioning size property, but would be a quick add.
- Or, if you want to keep the current button size, it might be worth fiddling with the styling to see if you can get the left and right margins to match up with the `.facet-list-element` (blue selection indicator thing). I think that might help, too.
- Range facet styling is also changing a bit, but the buttons used for those dropdowns are currently "btn-outline-dark" rather than "btn-outline-secondary." I'd normally ask you to switch to "btn-outline-dark" for the sake of consistency, but I do actually really like the way "btn-outline-secondary" looks here. @alexkb0009 Let me know if you have a preference between the two, too. If you also like "outline-secondary", I'll see if I can switch range facet to use that style in my current PR.

Bianca-Morris (commented):
Some small styling notes.

src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -405,7 +481,7 @@ export const CountIndicator = React.memo(function CountIndicator({ count = 1, co
     });
     return (
         <svg className="svg-count-indicator" viewBox={`0 0 ${width + 2} ${height + 2}`} width={width + 2} height={height + 2}>
-            { dots }
+            { dots}
```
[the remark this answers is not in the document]
alexkb0009: Haha
I guess one idea for us might be to setup/define code linting config file more, though on other hand it can be glitchy too or overly verbose/strict.

src/components/browse/components/FacetList/FacetTermsList.js
alexkb0009: Let's consider the style to converge to at the UI meeting. One consideration is how it looks from 4DN-side as well... though 4DN generally has "lighter" lines so pre-emptively I lean toward "btn-outline-secondary" a little bit as well due to that..

src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -284,18 +322,27 @@ export class FacetTermsList extends React.PureComponent {
...
-    'persistentCount' : 10
+    'persistentCount' : 10,
+    'defaultBasicSearchAutoDisplayThreshold': 15
 };
 
-
 const ListOfTerms = React.memo(function ListOfTerms(props){
-    const { facet, facetOpen, terms, persistentCount, onTermClick, expanded, onToggleExpanded, getTermStatus, termTransformFxn } = props;
+    const { facet, facetOpen, terms, onTermClick, expanded, onToggleExpanded, getTermStatus, termTransformFxn, searchText, onBasicTermSearch, onSaytTermSearch, persistentCount, defaultBasicSearchAutoDisplayThreshold } = props;
+    let { search_type: searchType = 'none' } = facet;
```
alexkb0009, line 337: (Low priority / doesn't matter too much) We could change this from `'none'` to `null`, benefit being that comparing null === null takes faster than 'none' === 'none' in the line below, but drawback being the change in type. I guess this subjective to large extent depending on how much prefer static types vs think the dynamic typing of JS is a benefit lol. I'm personally bit more in the latter camp, mostly because to maintain static types a good practice in my opinion might be to use something like TypeScript, but don't really want to take it on given other more-pressing priorities.

alexkb0009 (approved):
Looks & works awesome!

Pushed a minor somewhat unrelated commit to make the SAYT result rows more compact using [Bootstrap utility classNames](https://getbootstrap.com/docs/4.1/utilities/spacing/): [a](https://i.gyazo.com/55da39a827e5401ccd8abb561cd07887.png) -> [b](https://i.gyazo.com/49a25653496681f3df8f47816ba3337e.png).

PR #57, opened 2021-01-21
Range facet UI 2

Bianca-Morris:
Changelog:
- Updated to use outline-secondary bootstrap class https://gyazo.com/5262337e370256c70eeca821896c3d15 (Note, first two demos out of date in this regard -- still showing outline-dark)
- Updated w/ fix for animation smoothness https://gyazo.com/23f096f215fcb89636619b84800b8ef3
- Switched to bootstrap utility classes
- Added more padding around range-clear button when facet is closed to more closely match appearance of selected facets for stat facet lists
- Removed background color of range dropdown selections
- Fix: used "size" prop for bootstrap button size instead of adding new BS class to variant; removed old code (no more competing classes)
- Removed some extra CSS

Notes:
- Button widths aren't exactly, exactly the same, since text of "to" and "from" sizes are slightly different. I added a bit of margin to the second label to offset the first, so they're a little spaced out, but look closer to the same size.
- Please test in fourfront using [this branch](https://github.com/4dn-dcic/fourfront/pull/1422). I'll be doing some testing/tweaks for that today, as necessary.

Demos:
- [Style of Dropdown with range aggregation and without](https://gyazo.com/4109b5db3bf2350b6d337e8c02328242)
demo on padding

- [Better padding on closed selections](https://gyazo.com/877b7c0388a511c9531b669923d3305d)
demo on padding

- [Example of 4DN styling](https://gyazo.com/146066299f1cf85dbe35cee5c483442f)
demo on 4dn


alexkb0009 (approved):
Awesome!

(Off Topic / for later -)
In RangeClear in fallback for `const abbreviatedTitle = ..` we could change `.. || facetTitle;` to `.. || facetTitle.charAt(0).toUpperCase();` or maybe: `.. || (facetTitle && facetTitle.length > 6) ? facetTitle.charAt(0).toUpperCase() : facetTitle;` (?).

PR #58, opened 2021-01-22
Minor Updates/Simplification to SelectedItemsController

alexkb0009:
Some minor cleanup of the SelectedItemsController code. Not sure is fully complete, but I think ready to start being reviewed. I have **not** yet tested this on 4DN, which I think is only main place this being used still (for HiGlass file selection) (and not SubmissionView).

Functionality should stay exactly the same but hopefully code is slightly cleaner. Also a typo fix.

This is dependency of https://github.com/dbmi-bgm/cgap-portal/pull/300


src/components/browse/components/ControlsAndResults.js
```diff
@@ -83,7 +83,6 @@ export class ControlsAndResults extends React.PureComponent {
             // From ColumnCombiner or CustomColumnController (if props.hideColumns present):
             columnDefinitions,
             // From SelectedItemsController:
-            onCompleteSelection, onCancelSelection,
```
alexkb0009, line 86: Instead of defining these funcs in SelectedItemsController and passing them down into SelectStickyFooter, which just define the functions in SelectStickyFooter instead (since all the data for them is available in `selectedItems`).

src/components/browse/components/SelectedItemsController.js
```diff
@@ -129,42 +104,38 @@ export class SelectedItemsController extends React.PureComponent {
         // Kept for reference in case we want to re-introduce constrain that for 'select' button(s) to be visible in search result rows, there must be parent window.
         //var isThereParentWindow = inSelectionMode && typeof window !== 'undefined' && window.opener && window.opener.fourfront && window.opener !== window;
 
-        if (inSelectionMode) {
```
alexkb0009, line 132: This arrangement of `if` conditions is just a bit redundant, since we already return out if `!inSelectionMode`

Bianca-Morris (commented):
Similar to other PR, couldn't test this weekend due to eCommons auth issues. Code looks ok and I will test on Monday to double check for bugs.

Bianca-Morris (approved):

src/components/browse/components/SelectedItemsController.js
```diff
@@ -176,21 +147,46 @@ export class SelectedItemsController extends React.PureComponent {
...
 export const SelectStickyFooter = React.memo(function SelectStickyFooter(props){
-    const {
-        context, schemas, selectedItems,
-        onComplete, onCancel, currentAction
-    } = props;
+    const { context, schemas, selectedItems, currentAction } = props;
+
+
+    /** This function sends selected items to parent window. */
+    const onComplete = useCallback(function(){
```
alexkb0009, line 165: Moved definition of onComplete and onCancel into this `SelectStickyFooter` component in order to have a couple less props flowing through components (less code). Also it makes SelectedItemsController component more 'lean' and hopefully makes instances of reusage of it more clear. Both these functions only rely on `selectedItems` state of SelectedItemsController, and then the external/global `window`, so seemed like a safe move.

src/components/browse/components/SelectedItemsController.js
```diff
@@ -129,42 +104,38 @@ export class SelectedItemsController extends React.PureComponent {
...
-        const { children, ...propsToPass } = this.props;
+        const { children, ...remainingProps } = this.props;
         const { selectedItems } = this.state;
-
-        _.extend(propsToPass, {
+        const propsToPass = {
+            ...remainingProps,
             selectedItems,
             columnExtensionMap      : this.columnExtensionMapWithSelectButton(),
-            onSelectItem            : this.handleSelectItemClick,
```
alexkb0009, line 164: This was a typo -- corrected from `this.handleSelectItemClick` to `this.handleSelectItem`

utku-ozturk (approved):
I have tested it locally for both single and multiple selections; it works as it used to. The code also looks great!

alexkb0009:
Thanks for testing this out @utku-ozturk! (I was procrastinating on making a beta release for this lol)

PR #60, opened 2021-02-03
Rebuild tooltips when change hiddenColumns

alexkb0009:
No branches depend on this, just minor improvement. Can be glanced-over/reviewed/merged whenever have free time.


utku-ozturk (approved):
Tested locally, looks and works great!

PR #61, opened 2021-02-05
JWT updates

alexkb0009:
Complement to https://github.com/dbmi-bgm/cgap-portal/pull/319

Ignore any edits in `es/...` folder, these files are auto-generated


src/components/browse/components/VirtualHrefController.js
```diff
@@ -194,10 +194,6 @@ export class VirtualHrefController extends React.PureComponent {
                     throw new Error("Did not get back a search response, request was potentially aborted.");
                 }
 
-                if (typeof globalNavigate.updateUserInfo === "function") {
-                    globalNavigate.updateUserInfo();
-                }
```
alexkb0009, line 199: No longer needed here since is called after every AJAX request anyways via AJAXSettings.onLoad

src/components/util/ajax.js
```diff
@@ -99,11 +167,11 @@ export function promise(url, method = 'GET', headers = {}, data = null, cache =
             resolve(response);
         };
         xhr.onerror = reject;
-        if (cache === false && url.indexOf('format=json') > -1){
+        if (cache === false){
```
alexkb0009, line 170: This probably worth removing and then maybe setting Cache headers if cache is explicitly true. Idk. To consider later. The current version of this doesn't work anyways (might get "/some/path/&ts=..." (not "?"). Is unrelated to this branch so will save for later.

src/components/navigation/components/LoginController.js
```diff
@@ -63,87 +63,103 @@ export class LoginController extends React.PureComponent {
...
-        JWT.save(idToken); // We just got token from Auth0 so probably isn't outdated.
+    validateCookieAndObtainAdditionalUserInfo(token, successCallback = null, errorCallback = null){
+        const { updateAppSessionState, onLogin = null } = this.props;
 
         this.setState({ "isLoading" : true }, ()=>{
 
             this.lock.hide();
 
             // Second stage: get this valid OAuth account (Google or w/e) auth'd from our end.
+            // We probably can get rid of this Promise.race wrapper, since request/server will likely time out in 30s, idk..
```
willronchetti, line 118: if it's headed for timeout, you can bet on knowing about it after 30s.

src/components/navigation/components/LoginController.js
```diff
@@ -184,49 +202,73 @@ export class LoginController extends React.PureComponent {
...
+                        errorCallback(error);
+                    }
                 });
 
         });
+    }
 
+    auth0LoginCallback(authResult){
+        // First stage: we just have gotten JWT from the Auth0 widget but have not auth'd it against it our own system
+        // to see if this is a valid user account or some random person who just logged into their Google account.
```
willronchetti, line 215: While I do like inline comments like these, they aren't a replacement for docstrings IMO. The inline comments aren't always helpful if you don't understand how what it's doing fits into the overall scheme. This is more of a general comment as there are things you're doing to mitigate this ie: good function naming.

willronchetti (approved):
Approving with the caveat that I'm unlikely to catch JS specific errors. Generally the logic looks good.

src/components/navigation/components/LoginController.js
```diff
@@ -63,87 +63,103 @@ export class LoginController extends React.PureComponent {
...
 
     componentDidMount () {
         const { auth0ClientID, auth0Domain, auth0Options } = this.props;
-
+        // prefetch & preload enabled here since is likely that user might want to click Login very quickly after loading webpage.
         import(
             /* webpackChunkName: "auth0-lock-bundle" */
             /* webpackMode: "lazy" */
+            /* webpackPrefetch: true */
+            /* webpackPreload: true */
```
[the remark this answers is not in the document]
alexkb0009: It's supposed to queue up the [code-split-out auth0 library JS file to be loaded in a tiny bit earlier than previously](https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules). Not sure if it makes a practical difference since the auth0 library is queued to be loaded pretty early on anyway since is needed for the Login button to work.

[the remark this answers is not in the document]
alexkb0009: [Another relevant article](https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c)

utku-ozturk (approved):
tested locally in Firefox/Safari/Opera and Chrome, looks and works great! Btw, Chrome's behavior is slightly different compared to other browsers when cookies are modified manually in inspect panel: The user is logged out of the portal but can not log in unless they clear cookies from settings. Not sure, but this is a feature of Chrome for extra security. This is not the case for Firefox/Safari and Opera.

Here is a screencast:
https://t.gyazo.com/teams/ugurteam/9b31b617de1c57ba4aeb73379b71ebfb.mp4

Bianca-Morris (approved):
Once you merge, I'll try to update my ajax method for formdata posting to be integrated with this (I might end up merging it into the current `ajax.promise` function... but for now I'll use what I currently have). I don't think any of this will conflict with it (merge or otherwise). You can have a closer look when you review my branches.

src/components/util/analytics.js
```diff
@@ -139,9 +139,11 @@ export function initializeGoogleAnalytics(trackingID = null, appOptions = {}){
     ga2('create', trackingID, 'auto');
     ga2(function(tracker){
         const clientID = tracker.get('clientId');
+        console.log("Got client id", clientID);
         if (clientID){
             // Used on backend to associate downloads with user sessions when possible.
-            JWT.cookieStore.set('clientIdentifier', clientID, { path : '/' });
```
alexkb0009, line 144: We no longer have cookieStore (or the universal-cookie npm package, which has somewhat misleading documentation & functionality anyway); is simple enough to just do with raw JS. Maybe could set an expiry time as well, idk.

alexkb0009: Related - https://developer.mozilla.org/en-US/docs/Web/API/document/cookie

PR #63, opened 2021-02-26
SV fix for initialize fields w/value = 0 to null

Bianca-Morris:
Test with https://github.com/dbmi-bgm/cgap-portal/pull/339.

Changelog:
- Initialize zero values to zero instead of null in SubmissionView's BuildContext util (https://trello.com/c/6BWakdoQ/265-quick-edit-0-sometimes-shows-up-as-no-value-eg-loeuf)


alexkb0009 (approved):
Didn't test on local but edits look good & IMO good to merge if was working as expected on local.

PR #65, opened 2021-03-11
FilterSetUI + DetailPane related edits

alexkb0009:
Pulls out some code from RangeFacets to make reusable for FilterSetUI -- https://github.com/dbmi-bgm/cgap-portal/pull/360.
Also minor touchups to existing code.

### DIFF
--- a/scss/facet-list.scss
+++ b/scss/facet-list.scss
@@ -177,7 +177,8 @@ $facetlist-term-block-height: default !default;
                     
                     div.range-drop {
                         display: flex;
-                        align-items: center;                    
+                        align-items: center;  
+                        line-height: 1;
 
                         label {
                             display: inline-block;
--- a/scss/facet-list.scss
+++ b/scss/facet-list.scss
@@ -214,11 +214,16 @@ $facetlist-term-block-height: default !default;
                         .dropdown > .dropdown-menu {
                             .inline-input-container {
                                 padding-left: 10px;
-                                padding-bottom: 10px;
-                                border-bottom: 1px solid #ddd;
-                                margin-bottom: 8px;
                                 display: flex;
                                 align-items: center;
+                                margin-bottom: 0px !important; /* overrides #content form margin-bottom: 10px !important style. */
+
+                                &.has-options {
+                                    padding-bottom: 10px;
+                                    border-bottom: 1px solid #ddd;
+                                    margin-bottom: 8px !important;
+                                }
+
                                 > .input-element-container {
                                     flex: 1;
                                 }
--- a/src/components/browse/components/FacetList/RangeFacet.js
+++ b/src/components/browse/components/FacetList/RangeFacet.js
@@ -82,7 +82,7 @@ export function formatRangeVal(termTransformFxn, fieldFacetObj, rangeValue, allo
     }
 
     if (field_type === "date") {
-        return <LocalizedTime timestamp={value} localize={false} />;
+        return <LocalizedTime timestamp={rangeValue} localize={false} formatType="date-file" />;
     }
 
     if (field_type === "number"){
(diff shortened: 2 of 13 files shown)


scss/facet-list.scss
```diff
@@ -177,7 +177,8 @@ $facetlist-term-block-height: default !default;
                     
                     div.range-drop {
                         display: flex;
-                        align-items: center;                    
+                        align-items: center;  
+                        line-height: 1;
```
alexkb0009, line 181: Changes the implicit height of the 'From' and 'To' text areas so they center alongside the dropdown buttons a tiny bit better (IMO / possibly subjective).

Bianca-Morris: Maybe subjective, but I think Burak said something similar. I'm not always great with those details, so I'll take your word and his for it that this is an improvement.

src/components/browse/components/FacetList/RangeFacet.js
```diff
@@ -304,42 +349,18 @@ export class RangeFacet extends React.PureComponent {
...
-            if (absVal >= 1000) {
-                return decorateNumberWithCommas(transformedValue);
-            } else {
-                return transformedValue;
-            }
-        }
-
-        if (toPrecision) { return transformedValue.toPrecision(3); }
-        return transformedValue.toExponential(3);
+    termTitle(value, allowJSX = true){
```
Bianca-Morris, line 355: This docstring is confusing to me? fieldName no longer exists in this scope?

Bianca-Morris (commented):
Looks good. I did notice an attempted state update on unmounted component when selecting a new range which is probably worth looking into. https://gyazo.com/c98703889b1424a6bcd43732fc18ceaf It's possible that this was present before, but not sure.

Other than that, I think taking a second look at some of the docstrings/comments would be useful. Some are out of date or maybe just vague? Linked the main one I'm concerned about.

src/components/browse/components/SearchResultTable.js
```diff
@@ -393,12 +393,16 @@ class LoadMoreAsYouScroll extends React.Component {
         return styles;
     }
 
-    static getElementHeight(openDetailPanes, rowHeight, children, openRowHeight){
```
alexkb0009, line 396: No logic has changed here, just refactored a little bit for the logic to be more explicit/clear.

utku-ozturk (approved):
tested locally. works and looks great!

src/components/browse/components/FacetList/RangeFacet.js
```diff
@@ -70,12 +66,61 @@ export function getRangeValuesFromFiltersByField(facets = [], filters = []){
...
+            }
+        } else {
+            valToShow = valToShow.toPrecision(3);
+            // Try to prevent trailing 0s e.g. in 0.00000100
+            // Taken from https://stackoverflow.com/questions/26299160/using-regex-how-do-i-remove-the-trailing-zeros-from-a-decimal-number
+            valToShow = valToShow.replace(/(\.\d*?[1-9])0+$/g, "$1");
+        }
+    } // else is assumed to be valid JSX already
+
+    return valToShow;
```
Bianca-Morris, line 113: Looks like large decimal values are [no longer being rounded](https://gyazo.com/231252e80b4cb408079b96ce2831fee1). I remember going back and forth with Burak about this and he was pretty adamant about keeping those dropdown values rounded similarly to the items in the value indicator (around 3/4 significant digits). If there's a better/more consistent way than the `toExponential` method I was using (did notice some occasional weirdness with that: [from prod](https://gyazo.com/c7763575a6dff1228cd7bdaa616cd0f9)), let's do that (maybe we can just do `toPrecision(3 or 4)`, idk). 

Still not sure if this was his personal nitpick or something more general, but that was definitely something he noticed and got frustrated about every time I secretly changed it back. So maybe somewhat important for readability.

Bianca-Morris (approved):
Looks good to me, too. One note on rounding, though, shouldn't necessarily prevent a merge.

PR #66, opened 2021-04-07
Touchups related to FSUI + (fix:) Unset filtering state on FacetLists Terms

alexkb0009:
#### 2021-04-12
Cleanup

#### 2021-04-09
Minor update to CountIndicator (better determination of column idx when not 3 dots per column)

#### 2021-04-07

Bundled in fix for issue Will mentioned re: unsetting of FacetList Terms on /search/ and /browse/ views when click many (or single 1) rapidly. [Should now work like this](https://gyazo.com/a2e1aab36e2d27660c67ea8c34a2be8f).

#### 2021-04-06

Minor touchups done while working on FilterSetUI.

~~Despite the branch name, none of these are breaking changes nor does current FSUI (CGAP) PR depend on this. Can be merged whenever.~~
~~I take that back; added prop `ltr` to CountIndicator, which is used in CGAP FSUI now.~~
Nevermind again, `ltr` no longer used.

Most significant of these is auto-including institution and project in SubmissionView to allow creation to work. Not super pretty but it works.

### DIFF
--- a/es/components/browse/components/CustomColumnController.js
+++ b/es/components/browse/components/CustomColumnController.js
@@ -313,9 +313,9 @@ var ColumnOption = /*#__PURE__*/React.memo(function (props) {
 
   if (sameTitleColExists) {
     if (!description) {
-      showDescription = '<i class="icon icon-fw fas icon-code">&nbsp;</i><em class="text-300">' + field + '</em>';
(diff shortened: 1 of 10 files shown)


Bianca-Morris (approved):
Added a few minor notes. Thanks for that Institution + Project fix; I'd run into that a few times while testing various things -- very frustrating. It'll probably be important to test this one on 4DN, but definitely looks good from the CGAP side.

PR #67, opened 2021-04-08
fix for displayTitle onclick: use href from a elem

Bianca-Morris:
**Changelog:**
- Bugfix for search table onclicks: give priority for navigation to href from `<a>` element (target or nearest parent) instead of always using item atID. 
- Uses atID as a fallback in the case no target or parent `<a>` is available

**Notes:**
- Necessary for full functionality of https://github.com/dbmi-bgm/cgap-portal/pull/345
- Seems like if multiple `<a>` tags are present (really shouldn't happen unless someone wraps the output of the column render function in a `<a>`), it'll use the innermost `<a>`'s href. Best way to avoid is to not do that, and instead pass link/href as a prop to the render f(x) instead.


alexkb0009 (approved):
Looks good

PR #71, opened 2021-06-09
Item Details Updates for Multi-Dim Arrays and Attachments

utku-ozturk:
This PR solves the issues that explained in the tickets below:

1. 2D Array: https://trello.com/c/6CviAenq
2. Attachment: https://trello.com/c/dV3YiNdn

**SPC Release**: https://github.com/4dn-dcic/shared-portal-components/releases/tag/0.1.9b1

### 2D Array

The existing implementation renders 2D arrays as a table which is hardly readable. Instead of a specific solution only for 2D arrays, a general approach that renders multi-dim arrays having items of primitive data types (https://developer.mozilla.org/en-US/docs/Glossary/Primitive) as a list of tuples is preferred.

**Old:** 
af68625839c4b5222b9baee1676ad077

**New:**
Screen Shot 2021-06-09 at 00 07 24

### Attachment
Attachment links was not correctly rendered in tables, it is fixed.

**Old:**
Screen_Shot_2021-06-07_at_10 22 33_AM

**New:**
Screen_Shot_2021-06-08_at_16 36 22

### DIFF
--- a/src/components/util/misc.js
+++ b/src/components/util/misc.js
@@ -55,3 +55,7 @@ isSelectAction.selectionActionMap = {
     "multiselect" : true
 };
 
+export function isPrimitive(value) {
+    const t = typeof value;
+    return t === 'string' || t === 'number' || t === 'bigint' || t === 'boolean' || t === 'symbol' || t === 'undefined' || t === 'null';
+}
--- a/src/components/util/misc.js
+++ b/src/components/util/misc.js
@@ -57,5 +57,5 @@ isSelectAction.selectionActionMap = {
 
 export function isPrimitive(value) {
     const t = typeof value;
-    return t === 'string' || t === 'number' || t === 'bigint' || t === 'boolean' || t === 'symbol' || t === 'undefined' || t === 'null';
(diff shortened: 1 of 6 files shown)


src/components/util/misc.js
```diff
@@ -55,3 +55,7 @@ isSelectAction.selectionActionMap = {
     "multiselect" : true
 };
 
+export function isPrimitive(value) {
+    const t = typeof value;
+    return t === 'string' || t === 'number' || t === 'bigint' || t === 'boolean' || t === 'symbol' || t === 'undefined' || t === 'null';
```
alexkb0009, line 60: The 'type' of `null` is Object interestingly enough; if want to test for it might be better to do `t === null` ("is actual value of it null").

alexkb0009 (approved):
Code changes look decent, didn't test on local but if working as expected everywhere is good to merge IMO.

src/components/util/misc.js
utku-ozturk: @alexkb0009  Aha! Good catch, thanks. My bad, I implemented it while I was reading the MDN article (https://developer.mozilla.org/en-US/docs/Glossary/Primitive), and didn't notice the single quotes.

PR #73, opened 2021-06-16
FacetList: Prevent scrolling when ExtendedDescriptionPopover is open

alexkb0009:
(Also has some minor cleanup+comments.)


Bianca-Morris (approved):
Code looks good to me. I'll probably test tomorrow and let you know how it goes.

utku-ozturk (approved):
I was not able to test it on 4DN since extended_descriptions are not supported yet. As far as I understand, I should test the following behavior :) https://t.gyazo.com/teams/ff-4dn/033d1a8ebb20e7dc37ff57ddac519949.gif

The code looks and works great. IMHO, it is a good improvement in UX.

PR #74, opened 2021-06-17
Flexible Description Box Long Text Clipping Off Issue

utku-ozturk:
Please handle this PR along with https://github.com/4dn-dcic/fourfront/pull/1499

Trello: https://trello.com/c/Cypdy46K

Some lines of long descriptions are clipped off due to miscalculation of content height. This issue is observed in Chrome's desktop version and all mobile browsers.

Chrome has a known issue since version 47: https://stackoverflow.com/questions/65453475/getting-wrong-window-height-in-chrome

### Old:

Browser Chrome
Screen Shot 2021-06-16 at 11 55 23
Screen Shot 2021-06-16 at 11 53 09

### New:

BrowserChrome1
Screen Shot 2021-06-16 at 11 55 58
Screen Shot 2021-06-16 at 11 54 38

### DIFF
--- a/src/components/util/layout.js
+++ b/src/components/util/layout.js
@@ -224,7 +224,7 @@ export const textContentWidth = memoize(function(
 ){
     var contElem = document.createElement(containerElementType);
     contElem.className = "off-screen " + (containerClassName || '');
-    contElem.innerHTML = textContent;
+    contElem.innerHTML = textContent+ contElem.offsetHeight;
     if (style) contElem.style = style;
     contElem.style.whiteSpace = "nowrap";
     document.body.appendChild(contElem);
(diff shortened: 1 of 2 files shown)


src/components/util/layout.js
```diff
@@ -224,7 +224,7 @@ export const textContentWidth = memoize(function(
 ){
     var contElem = document.createElement(containerElementType);
     contElem.className = "off-screen " + (containerClassName || '');
-    contElem.innerHTML = textContent;
+    contElem.innerHTML = textContent + contElem.offsetHeight;
```
alexkb0009, line 227: Not sure this line is needed (I think it adds the value of height, like "138" or something, to end of textcontent, for purpose of the calculation), but yeah probably doesn't hurt to add some extra chars/padding to end. (Maybe could just use a comment that is happening; not too important though).

alexkb0009 (approved):

src/components/util/layout.js
utku-ozturk: @alexkb0009 Yes, I realized that the line was really unnecessary. I just asked @ugur-camoglu  whether there is a recent change since when we were testing, it was something different. (%99 sure)

PR #76, opened 2021-06-22
Support props.indeterminate in Checkbox

alexkb0009:
Checkbox component (which provides label wrapper) now uses `IndeterminateCheckbox` instead of `<input type="checkbox">` JSX element if `props.indeterminate` is passed in and set to `true`.

### DIFF
--- a/es/components/browse/components/CustomColumnController.js
+++ b/es/components/browse/components/CustomColumnController.js
@@ -130,30 +130,31 @@ export var CustomColumnController = /*#__PURE__*/function (_React$Component) {
   }, {
     key: "addHiddenColumn",
     value: function addHiddenColumn(field) {
-      this.setState(function (currState) {
-        if (currState.hiddenColumns[field] === true) {
+      this.setState(function (_ref) {
+        var currHiddenColumns = _ref.hiddenColumns;
+
+        if (currHiddenColumns[field] === true) {
           return null;
         }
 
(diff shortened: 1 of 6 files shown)


utku-ozturk (approved):
I have tested on 4DN, looks and works pretty good.

My only concern is (might be out of this PR's scope) indeterminate checkbox's onClick cycle is "`indeterminate - checked - unchecked - indeterminate - checked - unchecked ...`" in general, while the current cycle is "`indeterminate - checked - unchecked - checked - unchecked - checked - unchecked ...`". (it never turns back to indeterminate): https://t.gyazo.com/teams/ff-4dn/94d95a1ac46dd457ee53a237d5800cc3.gif

alexkb0009:
@utku-ozturk Thanks! Good catch, yeah that shouldn't happen (I think in that GIF, it should never be indeterminate to begin with), I'll double-check it.

utku-ozturk:
@alexkb0009 Sure, checkboxes in column listing options do not have/need `indeterminate` prop.  I added `indeterminate` to test it.

alexkb0009:
Re-implemented IndeterminateCheckbox using functional component w. callback ref. It seems possible the previous implementation stopped working with some browser or react version update given that it's cited as a solution years ago in this [StackOverflow thread](https://stackoverflow.com/questions/32139455/indeterminate-checkbox-in-react-jsx).

PR #77, opened 2021-06-29
hide doc-counts

Bianca-Morris:
See Dana's comment on this ticket: https://trello.com/c/zzuMrRXC/401-add-support-for-no-value-bucket-to-rangefacet

Test with: https://github.com/dbmi-bgm/cgap-portal/pull/415/ (symlinked) OR https://github.com/dbmi-bgm/cgap-portal/pull/427


Bianca-Morris:
Let me know if you think there's a better place/way to do this long term. This is basically a hotfix written quickly and for this very specific instance. Could use insight on if this might be better done in cgap-portal repo and then passed in via RangeFacet; was thinking about doing that, but knew this would be quicker to write/test.

alexkb0009 (approved):

PR #78, opened 2021-07-01
Performant Display of Stacked Tables Having Lots of Items

utku-ozturk:
Please handle this PR w/ https://github.com/4dn-dcic/fourfront/pull/1500

Trello: https://trello.com/c/oTpFC5hx

- Google Analytics lib does not respond well and makes the page crashing when there are too many files to be logged simultaneously. So we only track if the items are below a certain limit. (50)
- Instead of a collapsable panel that shows all items at once in stacked tables, we add an option to allow expanding the items in multiple steps. (https://t.gyazo.com/teams/ff-4dn/ab213aafe782e1536fa477c1cb0d7e3c.mp4) `incrementalExpandLimit` and `incrementalExpandStep` are introduced as new props.


alexkb0009 (approved):
Looks good!

(Didn't test on local)

PR #80, opened 2021-07-13
Sentry Integration

utku-ozturk:
Pls handle this PR with https://github.com/4dn-dcic/fourfront/pull/1501

Trello: https://trello.com/c/vAop8tlJ

resim


alexkb0009 (approved):
Looks great!

(for future branch/commit -)
Some errors/warnings may occur frequently-ish and maybe reasonably expected and could be excluded from sentry, but worth keeping all the logging right now to see how it works in prod and then cut down on some/any frequent/'ok' ones.

alexkb0009:
Another thing worth considering is maybe disabling Sentry for when working on localhost, or maybe when in development (to reduce amount of messages a bit). One idea is to check/use this `BUILDTYPE` injected variable which [gets added in by webpack](https://github.com/4dn-dcic/fourfront/blob/master/webpack.config.js#L106-L116). (It'd work same way as `SERVERSIDE` is used in isServerSide()).

src/components/util/logger.js
```diff
@@ -0,0 +1,100 @@
...
+export function info(message, ...arg) {
+    console.info(message, arg);
+    log(message, Sentry.Severity.Info, ...arg);
+}
+
+/**
+ * generic function to log into sentry
+ */
+function log(message, level, ...arg) {
+    Sentry.withScope(function (scope) {
```
alexkb0009, line 71: Should we check `isInitialized()` here before logging? Maybe revert to `console[level](message, arg)` if not isInitialized?

src/components/util/logger.js
```diff
@@ -0,0 +1,100 @@
+'use strict';
+
+import _ from 'underscore';
+import { isServerSide } from './misc';
+import * as Sentry from "@sentry/react";
+import { Integrations } from "@sentry/tracing";
+
+
```
alexkb0009, line 8: (Can create a global/module-level `let` variable or a `const` object/dictionary to preserve settings from initialization (i.e. `dsn`, `isInitialized`)).

src/components/util/logger.js
utku-ozturk: > Should we check `isInitialized()` here before logging? Maybe revert to `console[level](message, arg)` if not isInitialized?

@alexkb0009  `log()` function is called by `error()`, `warning()` and `info()` functions and not exported. So we added `isInitialized` check only to the `log()` function. If `isInitialized` is false, then the caller functions log into console and exit.

utku-ozturk:
> Looks great!
> 
> (for future branch/commit -)
> Some errors/warnings may occur frequently-ish and maybe reasonably expected and could be excluded from sentry, but worth keeping all the logging right now to see how it works in prod and then cut down on some/any frequent/'ok' ones.

Sure, let's monitor the frequencies of messages in the following days. Then we can revert some unnecessary log statements.

PR #81, opened 2021-08-05
Column's Initial Sort + Default Sort Fields

utku-ozturk:
Please handle this PR with https://github.com/4dn-dcic/fourfront/pull/1512

Trello: https://trello.com/c/6iOGV8dz


src/components/browse/components/table-commons/HeadersRow.js
```diff
@@ -120,6 +120,20 @@ export class HeadersRow extends React.PureComponent {
...
+            case 'string':
+                return 'asc';
+            case 'integer':
+                return 'desc';
+            case 'number':
+                return 'desc';
+            case 'date':
+                return 'desc';
+        }
+        return null;
```
alexkb0009, line 134: It'd be slightly more performant (due to speed of dictionary/object compared to switch-case - ~ O(1) vs O(N)) to do something like this -

```suggestion
        const directionsByFieldType = {
            "string": "asc",
            "integer": "desc",
            "number": "desc",
            "date": "desc",
        };
        return directionsByFieldType[fieldType] || null;
```

src/components/browse/components/table-commons/HeadersRow.js
```diff
@@ -199,16 +214,31 @@ export class HeadersRow extends React.PureComponent {
...
     sortByField(field){
-        const { sortColumns, sortBy } = this.props;
+        const { sortColumns, sortBy, columnDefinitions } = this.props;
         const [{ column = null, direction = "desc" } = {}] = sortColumns || [];
 
         const trimmedColumn = HeadersRow.getTrimmedColumn(column);
 
+        let initialSort;
+        if (columnDefinitions) {
+            const [colDef] = _.filter(columnDefinitions, function (item) { return item.field == field; });
```
alexkb0009, line 224: Use `find` or similar instead, it'll stop searching as soon as finds proper result instead of iterating over entire list (even if found matching result).

```suggestion
            const colDef = columnDefinitions.find(function (item) { return item.field == field; });
```

alexkb0009: Also instead of memoizing `getSortDirectionBySchemaFieldType`, would maybe create memoized func along lines of the following (to memoize the `.find`):
```
getInitialSort(columnDefinitions){
    const colDef = columnDefinitions.find(function (item) { return item.field == field; });
    if (colDef && (colDef.type || colDef.initial_sort)) {
        return colDef.initial_sort || getSortDirectionBySchemaFieldType(colDef.type);
    }
    return null;
}
```

alexkb0009 (approved):
Looks good, would maybe add the small changes suggested for slight performance boost. (Haven't loaded onto my local but assume has been tested).

es/components/browse/components/table-commons/HeadersRow.js
```diff
@@ -149,6 +149,25 @@ export var HeadersRow = /*#__PURE__*/function (_React$PureComponent) {
...
 
+  }, {
+    key: "getSortDirectionBySchemaFieldType",
+    value: function getSortDirectionBySchemaFieldType(fieldType) {
+      switch (fieldType) {
+        case 'string':
+          return 'asc';
+
+        case 'integer':
+          return 'desc';
```
andreacosolo, line 160: In my opinion the expected default order for integer and number should be ascending, whereas special cases like "% accuracy" could be overwritten as desc. But I am not familiar with CGAP schemas, maybe there the majority of numeric properties is the other way around?

src/components/browse/components/table-commons/HeadersRow.js
```diff
@@ -120,6 +120,16 @@ export class HeadersRow extends React.PureComponent {
         return column.substring(0, column.length - 14);
     });
 
+    static getSortDirectionBySchemaFieldType(fieldType) {
+        const directionsByFieldType = {
+            "string": "asc",
+            "integer": "desc",
```
andreacosolo, line 126: See my other comment above

es/components/browse/components/table-commons/HeadersRow.js
aschroed: @andreacosolo we may want to do a quick and informal review to see what makes most sense as default.  I'm thinking that often you do want number fields such as score or percent to initially show up desc and int fields could go either way - eg. replicate number or channel would want asc and number of experiments (if you can even sort on that field) desc - so a quick scan of columns in schema might provide guidance?  It may prove that CGAP and 4DN do want different default behavior.  As a first pass to get this change in I would propose leaving number DESC and change int to ASC but there may be important use cases so that the settings @utku-ozturk chose are sensible and schema changes can be made for fourfront.

aschroed (approved):
Please consider Andrea and my comments but I don't think these necessarily hold this PR back as it is a dependency for the related fourfront PR.

utku-ozturk:
> Please consider Andrea and my comments but I don't think these necessarily hold this PR back as it is a dependency for the related fourfront PR.

@aschroed @andreacosolo @alexkb0009  Thanks for your feedback. I updated the latest initial sort by data type as below and will merge it into master shortly.

```
const directionsByFieldType = {
            "string": "asc",
            "integer": "asc",
            "number": "desc",
            "date": "desc",
        };
```

PR #84, opened 2021-08-17
Sv UI 2 - Allow hiding multicolumn sort in AboveTableControlsBase/RightButtonsSection

Bianca-Morris:
**Changelog**
- Adds a prop for hiding `MultiColumnSort` button in `AboveTableControlsBase`.
  - Default behavior remains to show this by default, so shouldn't impact current implementations/usages of the component.
  - Default value is specified in `RightButtonsSection`
- Some random package-lock updates that generated post-symlink

Relevant CGAP PR: https://github.com/dbmi-bgm/cgap-portal/pull/460


utku-ozturk (approved):
Tested on local, works great!

alexkb0009 (approved):

PR #88, opened 2021-11-22
Minor button styling adjustments

alexkb0009:
- Changes some things in FacetList to be literal buttons (+ styling updates for it), such as the 'view more' buttons and facets' titles for better accessibility (i.e. ability to use keyboard interaction ([shift+]tab/enter/space) w. it).

- (No Effect) Adds 'fixed-height' class to some buttons which only have icon(s) as child (and no text); not used right now / may come back to later. On 4DN, there's a style rule in bootstrap_overrides.scss that sets >i:only-child to have line-height: inherit; this seems to work decently (but not perfectly; not used if 2+ icons in same button..) which could use some testing on CGAP-side.

### DIFF
--- a/es/components/browse/components/FacetList/FacetOfFacets.js
+++ b/es/components/browse/components/FacetList/FacetOfFacets.js
@@ -92,15 +92,16 @@ export var FacetOfFacets = /*#__PURE__*/function (_React$PureComponent) {
       return /*#__PURE__*/React.createElement("div", {
         className: "facet" + (facetOpen || anySelections ? ' open' : ' closed'),
         "data-group": title
-      }, /*#__PURE__*/React.createElement("h5", {
-        className: "facet-title",
+      }, /*#__PURE__*/React.createElement("button", {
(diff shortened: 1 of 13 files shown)


utku-ozturk:
I have tested locally and looks good to me! My only concern is that facet title looks darker and larger. I think, `h5.facet-title` to `button.facet-title` transition breaks the styling. Here is the old and new look and feel:

https://gyazo.com/39e54ea216a89ff92b7dbd5b8208df66

https://i.gyazo.com/7cc289c054bc5d864e4902d9ae91ef44.png

Bianca-Morris (approved):
Tested this on local CGAP, and it looks good (would be good to double check on 4DN re: Utku's concerns above).

I will say, though, that if the goal of making the switch to buttons is to get tab-navigation working, there's still a lot of work to be done in the FacetList component (there are some issues with how the dropdowns behave atm). Anyway, this is a step in the right direction, for sure.

PR #90, opened 2021-11-23
New Props for Facet Term's Icon Style and Persist Active Terms

utku-ozturk:
Pls handle this PR with its [FF pair](https://github.com/4dn-dcic/fourfront/pull/1560).

New props added to extend FacetList and Term components' styling:

- `useRadioButton`: Show either checkbox (False) or radio icon (True) for term component - it is only for styling, not intended to implement single selection (radio) or multiple selection (checkbox)
- `persistSelectedTerms`:  if True selected/omitted terms are escalated to top, otherwise each term is rendered in regular order. Moreover, inline search options are not displayed if it is False.

PS: [0.1.28b2](https://github.com/4dn-dcic/shared-portal-components/releases/tag/0.1.28b2) released for testing

Screen Shot 2021-11-23 at 22 16 16

### DIFF
--- a/src/components/browse/components/FacetList/FacetTermsList.js
+++ b/src/components/browse/components/FacetList/FacetTermsList.js
@@ -164,10 +164,12 @@ Term.propTypes = {
     'onClick'           : PropTypes.func.isRequired,
     'status'            : PropTypes.oneOf(["none", "selected", "omitted"]),
     'termTransformFxn'  : PropTypes.func,
-    'termIconStyle'     : PropTypes.oneOf(['check', 'radio'])
+    'termIconStyle'     : PropTypes.oneOf(['check', 'radio']),
+    'persistSelectedTerms' : PropTypes.bool.isRequired
 };
 Term.defaultProps = {
-    'termIconStyle': 'check'
+    'termIconStyle': 'check',
+    'persistSelectedTerms': true
 };
 
 /**
@@ -246,6 +248,7 @@ export class FacetTermsList extends React.PureComponent {
             filteringFieldTerm,
             setOpenPopover,
             termIconStyle,
+            persistSelectedTerms,
             context,
             schemas,
         } = this.props;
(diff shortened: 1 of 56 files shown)


src/components/browse/components/FacetList/FacetTermsList.js
```diff
@@ -117,18 +117,20 @@ export class Term extends React.PureComponent {
     }
 
     render() {
-        const { term, facet, status, termTransformFxn, isFiltering } = this.props;
+        const { term, facet, status, termTransformFxn, isFiltering, termIconStyle } = this.props;
+        
         const selected = (status !== 'none');
         const count = (term && term.doc_count) || 0;
+        const isRadio = termIconStyle === 'radio';
```
alexkb0009, line 124: Try to avoid passing down string props if can, in favor of boolean, if what we ultimately need is boolean (better performance in memoized components). Could probably simplify this and just have either `props.useRadioIcon` (w defaultProps.useRadioIcon=false).

alexkb0009:
Looks good, want to check on hotseat but then good to merge I think.
Would also glance @ PR https://github.com/4dn-dcic/shared-portal-components/pull/88 if have time, it affects couple of same files as touched here.

alexkb0009:
Also should update babel version to the latest on fourfront and local machines (or at least to [versions used on CGAP](https://github.com/dbmi-bgm/cgap-portal/blob/master/package.json#L57-L67)), then I think we would have less differences in /es/ compiled files.

alexkb0009 (approved):

src/components/browse/components/FacetList/FacetTermsList.js
utku-ozturk: @alexkb0009  Yes, definitely makes sense. Replaced `termIconStyle` w/ `useRadioButton`.

utku-ozturk:
> Also should update babel version to the latest on fourfront and local machines (or at least to [versions used on CGAP](https://github.com/dbmi-bgm/cgap-portal/blob/master/package.json#L57-L67)), then I think we would have less differences in /es/ compiled files.

Great! It reduced es files having diff from 48 to 29.

PR #102, opened 2022-03-03
React17 + Node16 related edits

alexkb0009:
**_Most_** of the changes are relatively cosmetic - changing propTypes and similar, because React17 warnings are more verbose.

**_However_** with Node v16 and NPM v8 there's a new syntax to `package-lock.json`, and because of that, it might stop working on Node v12 (current/existing version we use). We could consider removing package-lock.json from this repository using .gitignore and see if that would work in short-term, and/or wait to merge until can update 4DN to Node16+ as well. We should discuss.

---

Have also updated our fork of the react-infinite library. Its build system was out of date and had trouble to get it to build new compiled JS. So instead got rid of almost all of it and adopted the method to generate '/es/' build used in SPC (because this is ultimately all we need, don't need to build minified/uglified JS to be loaded directly by web-browser). It appears to work well/fine from what can tell.


alexkb0009:
Will push some more edits shortly to update some package.json versions

Bianca-Morris (approved):
These changes seem fine to me. But I do wonder from a development standpoint if we're losing anything that's quietly making our lives easier by getting rid of use strict? If it's too verbose, it's too verbose. But it might be good to try to keep track of any useful warnings/errors we're no longer getting and see if there are other ways of retaining those (via eslint, etc).

PR #103, opened 2022-03-10
Moment.js to Date-Fns Migration

utku-ozturk:
Trello: https://trello.com/c/eEhhxmcm

Draft release: [0.1.36b6](https://github.com/4dn-dcic/shared-portal-components/releases/tag/0.1.36b6)
FF PR: https://github.com/4dn-dcic/fourfront/pull/1617

Based on benchmarks and features, we decided to give [date-fns](https://github.com/date-fns/date-fns) library a chance.

### Features Comparison Table: 

features

### Parsing Performance

parse

### Get-Set Performance

getset


src/components/util/search-filters.js
```diff
@@ -33,11 +32,11 @@ export function getUnselectHrefIfSelectedFromResponseFilters(term, facet, filter
...
-            var interval = getDateHistogramIntervalFromFacet(facet) || 'month',
-                toDate = moment.utc(term.key);
-            toDate.add(1, interval + 's');
-            toFilter    = _.findWhere(filters, { 'field' : field + '.to',   'term' : toDate.format().slice(0,10) }),
-            fromFilter  = _.findWhere(filters, { 'field' : field + '.from', 'term' : term.key });
+            // var interval = getDateHistogramIntervalFromFacet(facet) || 'month',
+            //     toDate = moment.utc(term.key);
+            // toDate.add(1, interval + 's');
+            // toFilter    = _.findWhere(filters, { 'field' : field + '.to',   'term' : toDate.format().slice(0,10) }),
+            // fromFilter  = _.findWhere(filters, { 'field' : field + '.from', 'term' : term.key });
```
Bianca-Morris, line 39: Is this not actually in use anywhere at the moment? Just double checking it wasn't accidentally left commented out.

Bianca-Morris (commented):
Code changes seem good at a glance; I haven't gotten a chance to test this though, so I'll leave that to Alex. 

As a side note: I am curious what you used for benchmarking/profiling of the different potential libraries (I think you may have mentioned that during your presentation; hope you don't mind reminding me).

src/components/ui/LocalizedTime.js
```diff
@@ -147,21 +143,21 @@ export function preset(formatType = 'date-md', dateTimeSeparator = " "){
...
+export function display(dateObj, formatType = 'date-md', dateTimeSeparator = " ", localize = false, customOutputFormat = null){
     var outputFormat;
     if (customOutputFormat) {
         outputFormat = customOutputFormat;
     } else {
         outputFormat = preset(formatType, dateTimeSeparator);
     }
     if (localize){
-        return momentObj.local().format(outputFormat);
+        formatDate(dateObj, outputFormat, { locale: enUS });
```
alexkb0009, line 157: Can we have this auto-determine what locale the browser is in? We don't want to use enUS if user(s) are not from US.

src/components/util/search-filters.js
alexkb0009: We do have "date_created" field which is meant to be date_histogram, but don't think it's currently used as facet. This feature might not have been fully implemented in which case is worth removing or commenting out, but am not 100% sure. Would probably just comment out entire `if` block though so that it does `throw new Error('Histogram not currently supported.');` (line below) if `facet.aggregation_type === 'date_histogram'`.

alexkb0009: Just need to comment out line 34 -> `// } else if (facet.aggregation_type === 'date_histogram'){`

alexkb0009 (approved):
This should be all set, I just need to update CGAP to use date-fns as well before merging this.

PR #116, opened 2022-04-08
Dependency Updates

utku-ozturk:


package.json
```diff
@@ -61,6 +62,7 @@
         "aws-sdk": "^2.1096.0",
         "crypto-js": "^3.3.0",
         "date-fns-tz": "^1.3.0",
+        "date-fns": "^2.28.0",
```
alexkb0009, line 65: Why do we need these changes? `date-fns` is in peerDependencies already, and `date-fns-tz` is already in regular dependencies so adding it again to devDependencies is redundant.

utku-ozturk: @alexkb0009 As I talked to Ugur, 

- `date-fns-tz` is required for getting UTC values. (doc: https://date-fns.org/v2.28.0/docs/Time-Zones)
- `date-fns` is required in both dependencies block since linking is not working if it is missing.

PR #201, opened 2023-01-10
Child Popup Window Updates

utku-ozturk:
**Trello:** https://trello.com/c/dJCyoEJ0

Draft release: https://github.com/4dn-dcic/shared-portal-components/releases/tag/0.1.58b3

**Issues:**
Child popup windows are used for single or multiple item(s) selections in the data portal. One good example of selection would be adding files to the `HiGlassViewConfig`.

As is explained in detail in the Trello card, the initial page (SearchView) in the child window has a footer showing the count of the selected items along with Apply/Cancel buttons; but the footer disappears if you navigate to other pages. Since the child window lacks a top bar, you cannot quickly turn back to the initial page. (the only way is keyboard shortcuts or context menu)

Another issue is even if you manage to return to the selection page, you'll see that all previous selections you made are lost.

The desired behavior would be to navigate easily without losing previous selections.

**Solution:**
Since both 4DN and CGAP depend on SPC, the following additions are optional; they will only affect the current look and feel if you pass the proper prop values. (Please take a look at the [FF PR](https://github.com/4dn-dcic/fourfront/pull/1766) for usage in 4DN) 

1. `<BackNavigationStickyFooter />` component (new)

Screen Shot 2023-01-11 at 00 18 42

``` js
BackNavigationStickyFooter.propTypes = {
    'text': PropTypes.string,  // the text for the back button. Default value is 'Return to Selection List'.
    'tooltip': PropTypes.string, // the tooltip for the the back button. Default value is 'Go to selection page'.
    'navigateToInitialPage': PropTypes.bool //true: always returns to initial page, false: return to previous page. Default is true. 
};
```
2. New `keepSelectionInStorage` prop for `<SelectedItemsController />` component

`SelectedItemsController` is already in use for any selections in SearchView. The new `keepSelectionInStorage` prop makes storing selected items temporarily possible in localStorage (if available). Even if you navigate back-and-forths, selected items get restored when back to the selection page. The default value is false. In any case, `localStorage` get cleared right after child window is closed.

https://i.gyazo.com/7f2e1b0f51b9adab3e85820ebf221545.gif

### DIFF
--- a/es/components/browse/SearchView.js
+++ b/es/components/browse/SearchView.js
@@ -5,7 +5,7 @@ import _inherits from "@babel/runtime/helpers/inherits";
 import _possibleConstructorReturn from "@babel/runtime/helpers/possibleConstructorReturn";
 import _getPrototypeOf from "@babel/runtime/helpers/getPrototypeOf";
 import _defineProperty from "@babel/runtime/helpers/defineProperty";
-var _excluded = ["href", "context", "showClearFiltersButton", "schemas", "currentAction", "facets", "navigate", "columns", "columnExtensionMap", "placeholderReplacementFxn", "windowWidth"];
+var _excluded = ["href", "context", "showClearFiltersButton", "schemas", "currentAction", "facets", "navigate", "columns", "columnExtensionMap", "placeholderReplacementFxn", "keepSelectionInStorage", "windowWidth"];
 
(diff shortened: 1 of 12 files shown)


Bianca-Morris (approved):
Tested briefly on CGAP, and didn't see any issues; should be good. Still looking over the fourfront code. I'll update the review there when I'm done looking at it.

PR #206, opened 2023-01-18
Custom auth0 ui fix

Bianca-Morris:
**Changelog**
- Pull `auth0Client`, `auth0Domain`, and `auth0Options` from `/auth0_config` endpoint instead of static values or props
  - Note: `auth0Options` can still be passed to `<LoginController />` as a prop, and keys not included in the `/auth0_config` response will be used as fallback
- Some updates to generated ES files

**Relevant CGAP PR**:
https://github.com/dbmi-bgm/cgap-portal/pull/694/

**Deployment notes**:
This is currently deployed on [cgap-devtest](https://cgap-devtest.hms.harvard.edu/) with the above PR.

**Testing/Review Requests**:
It would be great if @utku-ozturk and/or @gsabanci could double-check that this still works as expected on fourfront, since I made some small edits since it was last worked on by Utku (thanks again for your help with this).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@hms-dbmi-bgm/shared-portal-components",
-    "version": "0.1.60b4",
+    "version": "0.1.60",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@hms-dbmi-bgm/shared-portal-components",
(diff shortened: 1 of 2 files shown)


dmichaels-harvard (approved):
LGTM

utku-ozturk (approved):
looks good to me!

PR #227, opened 2023-03-03
Not facets

Bianca-Morris:
**Changelog:**
- Pin React-Bootstrap version to 1.6.4 (due to API change)
- Allow for selecting new omitted terms in "exclusion mode"
  - Hide range and stat facets in exclusion mode
- Create a new component for displaying FontAwesome v6 icons as SVGs (temporarily - sort of hacky; should be used sparingly)
   - Some additional code for styling these; specifically calculating the appropriate fill color (may be overkill -- let me know if it is and I should simplify it) 
   - Note: icons would need to be added to the same folder in 4DN and CGAP for them to be used across both envs. See comments.
- Add a `stopPropagation` prop to `copyWrapper` component, so it can be placed inside some other things (needed for [this PR](https://github.com/dbmi-bgm/cgap-portal/pull/697))

**Relevant PRs:**
CGAP: https://github.com/dbmi-bgm/cgap-portal/pull/701
4DN: https://github.com/4dn-dcic/fourfront/pull/1789


utku-ozturk (approved):
tested on local, works great!

PR #229, opened 2023-03-09
auth0_lock package pinned to version 11

gsabanci:
https://trello.com/c/7zMQeEyI


Bianca-Morris (approved):
I tested this change while working with CGAP and it seemed fine.

I will say, the size of these ES files diffs is getting a bit unwieldy. Do you all know why this is happening? Is it the result of a difference in bundler or compiler package versions between CGAP and 4DN? I know you all upgraded to Webpack 5 somewhat recently, and we [haven't yet gotten around to that update](https://github.com/dbmi-bgm/cgap-portal/pull/574) (need to double check the status of this -- there were some issues with it that I don't think ever got resolved). If you have any insight, let me know; I'm hoping I'll have a little time to look into this when I get back.

PR #230, opened 2023-04-27
Add a LICENSE.txt

netsettler:
* Best practices is to have a `LICENSE.txt` with all of our repositories. I've added one here
* While I was here, I added some adjustments to `.gitignore` to know about a few more filenames.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,14 @@
+# npm
 node_modules
-.DS_store
+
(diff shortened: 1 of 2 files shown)


Bianca-Morris (approved):
Seems reasonable. Heads up to @utku-ozturk and @gsabanci that this is being added.

PR #236, opened 2023-08-01
Cfm link to items fix

crfmc:
Corresponds to [Trello Ticket](https://trello.com/c/UuNejTct/804-fix-for-linkto-items-in-nested-array-fields-on-itemdetailpane), and related to [cgap-portal pull request 732](https://github.com/dbmi-bgm/cgap-portal/pull/732).

- Adds conditional for nested objects that have "key" field of type string, as opposed to "childKeys" as an array. Fixes issues rendering


Bianca-Morris (approved):

Bianca-Morris (changes requested):
Actually just tested this on a few more items on devtest, and it seems like it's not working as expected:
- The tenth item in [this table](https://cgap-devtest.hms.harvard.edu/workflows/860a1240-f15a-425d-875b-7b8f943130d4/) of workflow arguments has a `secondary_file_format` of `vcf_gz_tbi`, which isn't displaying in the table
- The 11th item in that table also doesn't have a `secondary_file_format` but instead of displaying "-", is blank

Bianca-Morris (approved):
Okay! Looks good. Thanks for your patience with my mix-up!

PR #239, opened 2023-11-08
Smaht UI updates

Bianca-Morris:
### Changelog
- Add `CustomWrapper` prop to StaticPageBase
- Fix issue with rendering static pages and TOC caused by rename of `static_section.name` and `page.name` to `identifier` on SMaHT
- Additional support for `identifier` in `object.js` (probably not 100% necessary due to presence of `title`, but good to have)


crfmc (approved):

Bianca-Morris:
Will be merged with https://github.com/4dn-dcic/shared-portal-components/pull/243

PR #242, opened 2024-01-10
Cfm Item detail (QC) fix

crfmc:
Addresses Trello ticket [847](https://trello.com/c/tDD77iVi/847-bugfix-for-metaworkflows), [868](https://trello.com/c/lTO9fNBI/868-qc-value-bug), and [842](https://trello.com/c/73n6sPGX/842-truncated-fields-in-smaht-fix).

Features:
- Removes character limit for string values in SubItemTable
- For non-SubItemTable Input Argument fields, show argument names instead of numbers for dropdowns

Fixes:
- Value of 0 being typecast as false, and processed as non-existent
- Loss of nested information for Input Argument Values of type object

### DIFF
--- a/es/components/ui/ItemDetailList.js
+++ b/es/components/ui/ItemDetailList.js
@@ -176,10 +176,12 @@ var SubItemTable = /*#__PURE__*/function (_React$Component) {
         return _.map(columnKeys, function (colKeyContainer, colKeyIndex) {
           var colKey = colKeyContainer.key;
           var value = getNestedProperty(item, colKey);
-          if (!value) return {
-            'value': '-',
-            'key': colKey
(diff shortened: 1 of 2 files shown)


Bianca-Morris (approved):
Seems reasonable to me, I'm adding Utku as a reviewer since this is an SPC PR (you should also do this in the future).

utku-ozturk (approved):
Since MetaWorkflow is not supported in 4DN yet, I couldn't test those updates. I have tested the rest in local env. and it looks good to me.

Bianca-Morris:
Will be merged with https://github.com/4dn-dcic/shared-portal-components/pull/243

PR #243, opened 2024-01-19
Cfm QC and facet updates

crfmc:
This branch is a combination of branches [smaht-ui-updates](https://github.com/4dn-dcic/shared-portal-components/compare/smaht-ui-updates), [cfm-qc_fix](https://github.com/4dn-dcic/shared-portal-components/compare/cfm-qc_fix), and [cfm-facet_styles](https://github.com/4dn-dcic/shared-portal-components/compare/cfm-facet_styles).


utku-ozturk (approved):
not tested locally but looks good to me. (my only concern is the facet-list.scss updatea that we may need to override in 4DN but it is not a big deal)

Bianca-Morris (approved):

PR #244, opened 2024-01-25
Bm benchmarking table fix

Bianca-Morris:
### Changelog:
- Pass in `context` to search table header `colTitles` that are components
- Allow easier override of default search table cell alignment (default is still .text-center; you can just change it now)
   - Does not apply to null values, which are still always centered 
- Allow changes to `colTitle` in `columnExtensionMap` to take precedence over `columns` returned from backend (for `@type` only -- honestly wondering why this doesn't always happen, but don't want to risk breaking anything on 4DN or CGAP atm)
- Allow hiding tooltip for search table header cols via columnExtensionMap: `hideTooltip`
- Version bump to 0.1.74

### Related PRs:
https://github.com/smaht-dac/smaht-portal/pull/64

### SPC Pre-Release Version:
https://github.com/4dn-dcic/shared-portal-components/releases/tag/0.1.74b1


utku-ozturk (approved):
I think, column alignment is previously handled by render functions: https://github.com/4dn-dcic/fourfront/blob/fd31d5b420a523b1fc12095eb61c00d060cbf2a1/src/encoded/static/components/browse/columnExtensionMap.js#L213

`defaultColAlignment` prop is definitely more useful and simple.

crfmc (approved):
Looks good to me.

PR #247, opened 2024-02-23
feat: DotRoute/YoutubeVideo Embed and style updates

crfmc:
Contains SPC updates for three tasks:
- Style updates for SMaHT V2
    - move bootstrap styles into sass in order to overwrite them in parent repositories
- DotRouter component migration (from CGAP into SPC)
    - provide these components in SPC so they can be used in SMaHT and CGAP
- Style updates for item pages
    - make tables on item pages take up full width

Note: changes to padding in parent repos may need to be made


src/components/browse/components/above-table-controls/RightButtonsSection.js
```diff
@@ -29,7 +29,7 @@ export const ConfigureVisibleColumnsButton = React.memo(function ConfigureVisibl
     );
 });
 ConfigureVisibleColumnsButton.defaultProps = {
-    "className" : "btn btn-outline-primary"
+    "className" : "control visiblity btn"
```
utku-ozturk, line 32: possible typo for `visiblity?

crfmc: Thank you for catching this Utku! There is no use for this class at the moment actually, so I will remove.

Bianca-Morris:
@crfmc Is this in need of an update? Please take a second look and see if it should be closed or merged or updated. Thanks!

PR #257, opened 2024-06-24
feat: support separate maxHeights for facets and results tables

crfmc:
Corresponds to [Trello ticket 250](https://trello.com/c/pelxx3O1/250-smaht-embedded-table-scroll-functionality?filter=member:crfmc).
- Support separate maxHeights for facets and results tables


Bianca-Morris (approved):

src/components/browse/components/ControlsAndResults.js
```diff
@@ -75,8 +75,10 @@ export class ControlsAndResults extends React.PureComponent {
             isOwnPage = true,         // <- False when rendered by EmbeddedSearchView, else is true when from a SearchView
             isContextLoading = false, // <- Only applicable for EmbeddedSearchView, passed in by VirtualHrefController only, else is false always since we initialize immediately over search-response context that already has first 25 results
 
-            // From EmbeddedSearchView/manual-entry, used if isOwnPage is true
-            maxHeight = SearchResultTable.defaultProps.maxHeight,
+            // From EmbeddedSearchView/manual-entry, used if isOwnPage is true,
+            maxHeight,
+            maxFacetsBodyHeight = SearchResultTable.defaultProps.maxHeight,
```
utku-ozturk, line 80: @crfmc  It looks like `SearchResultTable.defaultProps.maxHeight `does not exist anymore. Is it ok that both `maxFacetsBodyHeight` and `maxResultsBodyHeight` are set as undefined?

utku-ozturk:
@crfmc I have tested it in 4DN locally. I think, we need to update static sections or pages that are using embedded search table with facets (we need to pass a prop?). For instance, when I run [this page](https://mastertest.4dnucleome.org/static-sections/4570c6b5-dad9-4f34-b366-9895b461334d/) in localhost, I am getting the client-side error below:

Screen Shot 2024-07-03 at 15 35 35

utku-ozturk:
Update: Adding maxHeight to defaultProps looks fixed it.

src/components/browse/components/ControlsAndResults.js
crfmc: Thank you for pointing this out! Yes I was running into the same error on CGAP, I am adding maxHeight back to defaultProps.

crfmc: Should work properly now on CGAP and Fourfront

utku-ozturk (approved):
Yeah, I've re-tested 0.1.86b3 locally, it looks good to me! thanks for the fix.

PR #262, opened 2024-09-06
Bug Fix - Table of Content props.children handling

utku-ozturk:
This fixes the "`props.children` used to be an array" assumption and converts non-array children to arrays before iterating through them.


Onurcankaratay (approved):

utku-ozturk:
[Bug - TOC generation fails for mixed types of static content (RST-MD.. etc.)](https://trello.com/c/q46uaJsd/342-bug-toc-generation-fails-for-mixed-types-of-static-content-rst-md-etc)

PR #263, opened 2024-12-17
Bm browse view

Bianca-Morris:
**Changelog:**
- Add a means of hiding the sticky footer that appears on SearchView when `currentAction` = `multiselect` (`hideStickyFooter` prop)
- Add a means of bypassing the use of `SelectedItemsController` in `SearchView` directly when `currentAction` = `multiselect` (`useCustomSelectionController` prop)
- Add an alternative layout to `AboveTableControlsBase` for SMaHT `BrowseView` table
  - Rename `RightButtonsSection` to `ColumnCustomizationButtons`, since they will no longer always be on the right side
  - Additional customization for `ColumnCustomizationButtons` (including button child styling, rendering without wrapper div)
- Rework `bytesToLargerUnit` to allow showing only unit or only value

**Related SMaHT PR:**
- https://github.com/smaht-dac/smaht-portal/pull/305

**Related Trello Ticket:**
- https://trello.com/c/kHZnft8v/370-production-data-table-v1-implementation


utku-ozturk:
@Bianca-Morris FYI, I pushed two minor commits to make it work with 4DN. (otherwise, it doesn't render the button styles)

Screen Shot 2024-12-24 at 13 48 22

PR #266, opened 2025-03-25
LocalizedTime UTC Updates

utku-ozturk:
Neutralize `date-fns'`s UTC adjustments for date time input lacks timezone.

For instance, if we have string literal `2025-01-01` as input and need to display it as `2025-01-01` using `LocalizedTime`, it is displayed as `2024-12-31` in GMT- time zones, whereas in GMT+ time zones it remains `2025-01-01`. Even though the represented datetime object is correct, the displayed date misleads users, particularly in facet date range components.


utku-ozturk:
[Facet date range picker is not working correctly for all time-zones. Adjust it for UTC.](https://trello.com/c/szAHOhFA/653-facet-date-range-picker-is-not-working-correctly-for-all-time-zones-adjust-it-for-utc)

crfmc (approved):

